PR #12, opened 2024-02-08
withDataService UpdateAll method

ricochetbrown:
We have a need for an updateAll method in withDataService as we handle archiving inside of SQL. We would like to pass all the entities in certain cases. This doesn't necessarily need to be part of the interface, but would like the method to be available.


rainerhahnekamp (approved):
Looks good to me. Nothing from the existing code is changed, only new features have been added.

rainerhahnekamp:
@ricochetbrown `updateAll` is now available in 0.0.4

PR #16, opened 2024-02-27
misc: conditional type narrowing on set callstate

andersfgraneng:
When trying to use `withCallState` in a `signalStore` and using `patchState` with the setter-functions, typescript will complain unless your store satisfies both `NamedCallStateSlice<Prop>` and `CallStateSlice` even if you only really get one of the two objects back. With this you should get conditional typing based on the input and not the union type of `NamedCallStateSlice<Prop>` and `CallStateSlice`.

E.g `setLoading('flights')` should give the type `NamedCallStateSlice<'flights'>` and not `NamedCallStateSlice<'flights'> | CallStateSlice`.

I appreciate any response as I think this would be nice if using `withCallState` for requests without `withDataService`.


rainerhahnekamp:
@andersfgraneng, thanks for letting us know and the PR. we'll look into it.

PR #17, opened 2024-02-28
Build pipeline stabilization

bohoffi:
This PR targets some build pipeline issues:
* demo application budgets
* the libs dependencies according to the dependency check Lint rule
* unified usage of shared `Empty` type
  * type was renamed

### DIFF
--- a/apps/demo/project.json
+++ b/apps/demo/project.json
@@ -41,8 +41,8 @@
           "budgets": [
             {
               "type": "initial",
-              "maximumWarning": "1mb",
-              "maximumError": "1.5mb"
+              "maximumWarning": "500kb",
+              "maximumError": "1mb"
             },
             {
               "type": "anyComponentStyle",
--- a/apps/demo/project.json
+++ b/apps/demo/project.json
@@ -41,8 +41,8 @@
           "budgets": [
             {
               "type": "initial",
-              "maximumWarning": "500kb",
-              "maximumError": "1mb"
+              "maximumWarning": "1mb",
+              "maximumError": "1.5mb"
             },
             {
               "type": "anyComponentStyle",
--- a/libs/ngrx-toolkit/package.json
+++ b/libs/ngrx-toolkit/package.json
@@ -9,10 +9,9 @@
   "peerDependencies": {
     "@angular/common": "^17.0.0",
     "@angular/core": "^17.0.0",
-    "@ngrx/signals": "^17.0.0"
-  },
-  "dependencies": {
-    "tslib": "^2.3.0"
+    "@ngrx/signals": "^17.0.0",
+    "@ngrx/store": "^17.0.0",
+    "rxjs": "^6.5.3 || ^7.5.0"
   },
   "sideEffects": false
 }
--- a/libs/ngrx-toolkit/package.json
+++ b/libs/ngrx-toolkit/package.json
@@ -9,9 +9,10 @@
   "peerDependencies": {
     "@angular/common": "^17.0.0",
     "@angular/core": "^17.0.0",
-    "@ngrx/signals": "^17.0.0",
-    "@ngrx/store": "^17.0.0",
-    "rxjs": "^6.5.3 || ^7.5.0"
+    "@ngrx/signals": "^17.0.0"
+  },
+  "dependencies": {
+    "tslib": "^2.3.0"
   },
   "sideEffects": false
 }
--- a/libs/ngrx-toolkit/src/lib/shared/empty.ts
+++ b/libs/ngrx-toolkit/src/lib/shared/empty.ts
@@ -1,2 +1,2 @@
 // eslint-disable-next-line @typescript-eslint/ban-types
-export type Emtpy = {};
+export type Empty = {};
--- a/libs/ngrx-toolkit/src/lib/shared/empty.ts
+++ b/libs/ngrx-toolkit/src/lib/shared/empty.ts
@@ -1,2 +1,2 @@
 // eslint-disable-next-line @typescript-eslint/ban-types
-export type Empty = {};
+export type Emtpy = {};
--- a/libs/ngrx-toolkit/src/lib/with-call-state.ts
+++ b/libs/ngrx-toolkit/src/lib/with-call-state.ts
@@ -5,7 +5,7 @@ import {
   withComputed,
   withState,
 } from '@ngrx/signals';
-import { Empty } from './shared/empty';
+import { Emtpy } from './shared/empty';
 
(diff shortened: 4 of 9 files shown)


apps/demo/project.json
```diff
@@ -41,8 +41,8 @@
           "budgets": [
             {
               "type": "initial",
-              "maximumWarning": "500kb",
```
rainerhahnekamp, line 44: Can we keep the budget with the default values? Can we keep below the threshold by layz loading routes? I think we should try to fix our code before touching the default settings?

libs/ngrx-toolkit/package.json
```diff
@@ -9,10 +9,9 @@
   "peerDependencies": {
     "@angular/common": "^17.0.0",
     "@angular/core": "^17.0.0",
-    "@ngrx/signals": "^17.0.0"
-  },
-  "dependencies": {
```
rainerhahnekamp, line 14: I think that nx might add this one automatically upon build? Can you check?
Here's more information: https://github.com/softarc-consulting/sheriff/commit/6f80696f9b6d8d6e76af40e445afa661180cf495

libs/ngrx-toolkit/src/lib/shared/empty.ts
```diff
@@ -1,2 +1,2 @@
 // eslint-disable-next-line @typescript-eslint/ban-types
-export type Emtpy = {};
```
rainerhahnekamp, line 2: I am not sure about that to be honest. I know you just fixed a typo there, but do we really want to call it empty. I can remember of some eslint rule which says `Empty` should be `Record<string, never>`.

libs/ngrx-toolkit/src/lib/with-call-state.ts
```diff
@@ -14,27 +14,27 @@ export type NamedCallStateSlice<Collection extends string> = {
 };
 
 export type CallStateSlice = {
-  callState: CallState
```
rainerhahnekamp, line 17: Can you maybe find a way of not reformatting those files? We might have to come up with a separate PR which runs prettier with a default configuration over all files but maybe this one is not suited for it.

There are so many changes and I have a hard time to differentiate reformatting and actual modifications.

libs/ngrx-toolkit/package.json
```diff
@@ -9,10 +9,9 @@
...
-    "tslib": "^2.3.0"
+    "@ngrx/signals": "^17.0.0",
+    "@ngrx/store": "^17.0.0",
```
rainerhahnekamp, line 13: Can we leave that out for now? @mikezks, I think we need to find a better solution here. It shouldn't be that ngrx/store (and rxjs) come as dependency of the toolkit.

rainerhahnekamp (commented):
Thanks @bohoffi, I've left some comments.

libs/ngrx-toolkit/package.json
mikezks: Fully agree. I would not put store and signals into dependencies or peerDependencies.

I may put parts of the Redux Connector into a secondary entry point later to add a peer dependency on the store, but for now I would remove them again.

libs/ngrx-toolkit/package.json
bohoffi: There was once a setting to keep dependencies up to date during build but it was removed in a [v17 migration](https://github.com/nrwl/nx/blob/master/packages/js/src/migrations/update-17-0-0/remove-deprecated-build-options.ts#L28).

The `@nx/dependency-checks` rule provides fixers for missing or unused but defined dependencies - so it could be cleaned up automatically.

libs/ngrx-toolkit/package.json
bohoffi: The `@nx/dependency-checks` rule which brought up those changes offers fixers for missing or unused dependencies - the result would be exactly the same as with this PR.

With Mikes potential plan ahead I'll revert this change.

libs/ngrx-toolkit/src/lib/shared/empty.ts
bohoffi: Yeah, eslint itself recommends `Record<string, never>`:
> If you want a type meaning "empty object", you probably want `Record<string, never>` instead.'

What do you think about renaming the type to `EmptyObject` to create a better alignment between its name and its usage and use `Record<string, never>` to apply eslints recommendation?

libs/ngrx-toolkit/src/lib/with-call-state.ts
bohoffi: I've reverted the changes and deactivated prettier for the workspace for now. I haven't thought about my autoformat on save setting after applying the rename for the `Empty` type.

apps/demo/project.json
bohoffi: Lazy loading the routes worked using the default settings

libs/ngrx-toolkit/package.json
rainerhahnekamp: Okay, then we might check that in another issue. Let's leave it for now.

libs/ngrx-toolkit/src/lib/shared/empty.ts
rainerhahnekamp: Absolutely, would you actually create an alias for that or go directly with the `{}`?

I heard once that TypeScript is faster, when using aliases instead of "individual" types.

libs/ngrx-toolkit/package.json
rainerhahnekamp: @mikezks does that actually work? 

https://github.com/testing-library/angular-testing-library/tree/main/projects/testing-library/jest-utils has the same issue with a dependency to Jest. The fixed it by not including it into the `package.json` (like we have it now).

apps/demo/project.json
rainerhahnekamp: Great!

libs/ngrx-toolkit/src/lib/shared/empty.ts
bohoffi: Personally I prefer a type alias as it ensures a consistent code style.
I've just checked what would happen when we'd use `Record<string, never>` instead of `{}` and it would prevent state patching as expected as the Record type wouldn't allow any updates.
Maybe I'm trying to overengineer it right now... when looking at its usage it represents that either do not have any specific expectations to the store or do not add anything to state/signals/methods.

I did some research about the performance and it depends - aliases can bloat up the bundle when massively used and inlined.

rainerhahnekamp:
I think the proposed changes have already landed. Therefore closing this PR. @bohoffi feel free to open it, if you think it is worth adding it.

PR #23, opened 2024-03-05
misc: refactored named types of withCallState

flensrocker:
This is a suggestion for an easier mapping of the state and signal types to the named (collection) types.

If you're ok with this I can take a look at the other mappings.

I've also reordered them so that the dependent type comes first.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-call-state.ts
+++ b/libs/ngrx-toolkit/src/lib/with-call-state.ts
@@ -9,26 +9,22 @@ import { Emtpy } from './shared/empty';
 
 export type CallState = 'init' | 'loading' | 'loaded' | { error: string };
 
-export type NamedCallStateSlice<Collection extends string> = {
-  [K in Collection as `${K}CallState`]: CallState;
-};
-
 export type CallStateSlice = {
   callState: CallState
 }
 
-export type NamedCallStateSignals<Prop extends string> = {
-  [K in Prop as `${K}Loading`]: Signal<boolean>;
-} & {
-    [K in Prop as `${K}Loaded`]: Signal<boolean>;
-  } & {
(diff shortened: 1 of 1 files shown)


rainerhahnekamp:
Hi Lars, thanks for the PR. We are more than just OK, if you find look for improvements.

libs/ngrx-toolkit/src/lib/with-call-state.ts
```diff
@@ -9,28 +9,24 @@ import { Emtpy } from './shared/empty';
...
-  } & {
-    [K in Prop as `${K}Error`]: Signal<string | null>;
-  }
+export type NamedCallStateSlice<Collection extends string> = {
+  [K in keyof CallStateSlice as `${Collection}${Capitalize<K>}`]: CallStateSlice[K];
+}
 
 export type CallStateSignals = {
   loading: Signal<boolean>;
   loaded: Signal<boolean>;
```
the-ult, line 22: Could it be an idea to rename `loaded` to `success` or something like that?
Since the state of the response is `successfull`. Otherwise it would be `error`. Correct?

Might be a bigger (breaking) change.. And therefore needs a different PR if @rainerhahnekamp agrees.
In that case, I could provide a PR with the proposed change

flensrocker: This PR is just an improvement in typing. Renaming properties (which is a breaking change) should be another PR.

rainerhahnekamp: @the-ult, I agree to @flensrocker. I see this PR as an improvement (refactoring). If we want to bring up new features/breaking changes, then please let's open another issue and discuss it there

rainerhahnekamp:
Thanks Lars!

PR #26, opened 2024-03-14
withDataService: partial entity update

qooco:
This is a changes for DataService methods (create & update & updateAll) to use Partial Entity 

---

Also have question, why update method receives only one argument Entity, 
```
update(entity: Partial<Entity>)
```
is not better?
```
update(id: EntityId, data: Partial<Entity>)
```


qooco:
A bit busy, no time.

rainerhahnekamp:
Thanks @qooco, the `update` didn't require the id so far, because it wasn't a partial. We'll discuss it internally. I support your changes. The only exception is the create. There, I don't want to have a partial.

qooco:
Ok I Understand about update.

About create, thats too ok, in most cases it is right I think (except such properties as created_at|updated_at). 

So thanks, good luck

qooco:
PR update: only Partial<E> update

PR #31, opened 2024-04-06
serializing bigint

robmanganelly:
Content of this PR:
- 2 Functions that wrap JSON.stringify and JSON.parse to include serialization and rehydration of bigint values. 
- Unit test code modifications to include values of type bigint on test cases.

Considerations.
- I applied the functions in place for the sake of brevity, but next iterations could move them into utility space and get them their own unit test.
- There is a commented test case, see line 137 that has failing test case, which i did not fix, as it is a different issue, unrelated to this PR.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-storage-sync.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-storage-sync.spec.ts
@@ -1,14 +1,16 @@
 import { getState, patchState, signalStore, withState } from '@ngrx/signals';
-import { withStorageSync } from './with-storage-sync';
+import { withBigIntRehydration, withBigIntReplacement, withStorageSync } from './with-storage-sync';
 import { TestBed } from '@angular/core/testing';
 
 interface StateObject {
   foo: string;
   age: number;
+  big: bigint;
 }
 const initialState: StateObject = {
   foo: 'bar',
   age: 18,
+  big: BigInt(123) //can't use bigint literal while target < 2020, see tsconfig.base.jsonm
 };
 const key = 'FooBar';
 
@@ -36,9 +38,10 @@ describe('withStorageSync', () => {
       // prefill storage
       localStorage.setItem(
         key,
-        JSON.stringify({
+        withBigIntReplacement({
           foo: 'baz',
           age: 99,
+          big: BigInt(123),
         } as StateObject)
       );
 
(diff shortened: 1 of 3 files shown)


libs/ngrx-toolkit/src/lib/with-storage-sync.spec.ts
```diff
@@ -1,14 +1,16 @@
 import { getState, patchState, signalStore, withState } from '@ngrx/signals';
-import { withStorageSync } from './with-storage-sync';
+import { withBigIntRehydration, withBigIntReplacement, withStorageSync } from './with-storage-sync';
```
rainerhahnekamp, line 2: Would it be possible to rename `withBigIntRehydration` and `withBigIntReplacement` to `serialize` and `deserialize`. I think `bigint` will not be the last type where we require custom code.

libs/ngrx-toolkit/src/lib/with-storage-sync.spec.ts
```diff
@@ -1,14 +1,16 @@
...
 import { TestBed } from '@angular/core/testing';
 
 interface StateObject {
   foo: string;
   age: number;
+  big: bigint;
```
rainerhahnekamp, line 8: Would it be possible, that we leave the tests as they are and just create a separate test which checks the de/serialization for `bigint`?

package-lock.json
rainerhahnekamp, line 1: I don't think that the `package-lock.json` should be changed, right?

libs/ngrx-toolkit/src/lib/with-storage-sync.ts
```diff
@@ -58,15 +104,20 @@ export type SyncConfig<State> = {
   /**
    * Function used to parse the state coming from storage.
    *
-   * `JSON.parse()` by default
+   * ~~`JSON.parse()` by default~~
```
rainerhahnekamp, line 107: Yes, you can replace that with `parse` and `stringify` (below).

rainerhahnekamp (changes requested):
Good work @robmanganelly. I've left some comments.

robmanganelly:
Sorry for the late reply. Yes. I will take care of it. Thanks for the
comments.

rainerhahnekamp:
@robmanganelly if you don't have the time, no worries. We'll merge and I can do the changes myself. Just let me know, cheers!

robmanganelly:
Honestly, I don't want to pass the workload onto you guys. I'm really happy
to contribute to this project, even if my contributions are minuscule, but
I currently don't have the bandwidth, sadly, for which I apologize.
 I can reach out once I'm done with my current tasks, and if changes are
still needed, I'll be more than happy to handle them myself.

Otherwise, please proceed as you see fit.

PR #48, opened 2024-05-13
Add a form management feature to a signal store

ebondu:


rainerhahnekamp:
@ebondu thanks for the PR. Would it be possible for you add a component which uses that feature in the showcase app?

ebondu:
I can try to add a component to demonstrate the feature but actually the showcase app doesn't compile. I tried to disable strict mod in tsconfig without success. Any idea ?

rainerhahnekamp:
Fixed the build of the demo app. Please rebase and try again

ebondu:
I added a small exemple in the demo app you should be able to try

PR #50, opened 2024-05-15
Added getSignalValues along with a demo

kobi2294:


rainerhahnekamp:
Hey Kobi, thanks, that's great. I haven't check it out so far (although I know what it is from our conversation). Would it be possible that you also your feature to the documentation?

rainerhahnekamp:
Hey @kobi2294, since this PR is already old and we have to clean up things here, I am going to close this one.

That doesn't mean that it is not accepted anymore, if the need still persists. You just have to rebase your branch and reopen this PR.

---

I am not sure about the use case anymore. Was it about getting one Signal that represents the whole state? Have you checked `computed(() => getState(store))`? Would that work?

PR #52, opened 2024-05-27
chore: depend only on @ngrx/signals

### LINKED ISSUE #51: Support for Angular 18
I would like to request support for Angular 18 in the NgRx Toolkit. 

Currently, I am unable to upgrade to Angular 18 because I rely on the NgRx Toolkit, which does not support this version yet.

rainerhahnekamp:
removing peerDependencies for angular
because they are transitive

fixes #51


rainerhahnekamp:
Not sure, if I get away with removing Angular as dependency, but let's see. Nx Release will have the last word in this.

PR #54, opened 2024-06-07
feat: Add local pagination feature

wolfmanfx:
This commit adds the implementation for local pagination in the `withPagination` function. It allows for displaying a subset of items from a larger collection, based on the current page and page size.

The `withPagination` function accepts an optional `collection` parameter to support multiple paginated collections. It calculates the selected page entities, total count, total pages, and page navigation array based on the provided options.

Additionally, utility functions like `gotoPage`, `setPageSize`, `nextPage`, `previousPage`, `firstPage`, and `setMaxPageNavigationArrayItems` are provided to easily update the pagination state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@angular/platform-server": "~17.0.0",
     "@angular/router": "~17.0.0",
     "@angular/ssr": "~17.0.0",
-    "@ngrx/signals": "^17.1.0",
+    "@ngrx/signals": "^17.2.0",
     "@ngrx/store": "^17.1.0",
     "@nx/angular": "18.3.3",
     "express": "~4.18.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "@angular/platform-server": "~17.0.0",
     "@angular/router": "~17.0.0",
     "@angular/ssr": "~17.0.0",
-    "@ngrx/signals": "^17.1.0",
+    "@ngrx/signals": "^17.2.0",
```
rainerhahnekamp, line 24: Is that mandatory?

wolfmanfx: Yes I had some type problems with the withEntity when I used 17.1.0

PR #58, opened 2024-06-12
fix(undo-redo): prevent duplicated entries on undo stack

manfredsteyer:
Perform a deep comparison to prevent duplicate entries on the stack. This can, for example, happen after an undo if the component sends back the undone filter to the store.


rainerhahnekamp:
@manfredsteyer do you maybe have a small snippet?

manfredsteyer:
@rainerhahnekamp What do you mean with small snippet?

rainerhahnekamp:
Just some pseudo code so that I can come up with a unit test for the 

>  after an undo if the component sends back the undone filter to the store.

I'm not sure if I fully understand it.

PR #72, opened 2024-07-24
fix: failing ci because of npm i

rainerhahnekamp:


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         with:
           node-version: '18'
           cache: 'npm'
```
jits, line 17: It's worth changing this to `cache: 'yarn'` as a result of the changes in this PR.

(Sorry for the drive-by commenting, instead of doing a PR)

rainerhahnekamp:
Hey @jits, you can edit it directly via GitHub: https://github.com/angular-architects/ngrx-toolkit/blob/main/.github/workflows/build.yml and open the PR.

I can do it by myself, but I'd like to have you as contributor

jits:
@rainerhahnekamp — ahh yes, great point! Will do that now.

jits:
@rainerhahnekamp — (for posterity) looks like that will create a fork, which makes sense.

I wonder though, would you consider moving to `pnpm` instead? I've had really good experiences using `pnpm` (example: https://github.com/FullStacksDev/angular-and-firebase-template/blob/main/.github/workflows/ci.yml#L32-L42). If so, I'd be happy to contribute a PR to switch this repo to `pnpm`.

rainerhahnekamp:
I tried pnpm several years ago and failed miserably. I went for yarn because of its stability and it is also used in the larger projects (angular, ngrx, etc) What advantages does pnpm have over yarn?

jits:
I find pnpm _smoother_ (gut feel), after many years of using yarn (and I've switched over to pnpm for all Node.js based projects now).

A few concrete reasons why this might be the case:

- **performance** — because pnpm uses symlinks to point your dependencies to a single source of truth on disk, it's very fast. 
- **better dependency isolation** — pnpm doesn't flatten the dependencies in the `node_modules` folder, so (in theory) provides a safer isolation between conflicting versions of sub-dependencies. It will still report conflicting versions though (so you can take action). My experience has been smoother in these situations with pnpm than yarn, where the latter would decide for you which version of the sub-dependency should be installed.
- **support** — Yarn v1 (which is what `npx yarn` would fetch and use) has _kinda_ been deprecated / is unsupported (https://github.com/yarnpkg/yarn?tab=readme-ov-file#%E2%84%B9%EF%B8%8F-important-note). It's not clear to me how long of a future it has. And the compatibility and migration story between v1 and subsequent versions isn't particularly smooth (https://yarnpkg.com/migration/guide). I've had annoying experiences migrating from Yarn v1 to a later version.
  - Whereas pnpm is gaining a lot of traction — e.g. it's the preferred choice in the Nuxt and unjs ecosystem.

I think I had a similar experience to you a few years back, but now it seems pnpm has matured quite a bit.

(https://refine.dev/blog/pnpm-vs-npm-and-yarn/ is a more detailed comparison, if you want to dig further)

PR #74, opened 2024-07-25
Switch to pnpm (for local dev)

### LINKED ISSUE #73: Moving to pnpm
Resulted from the following discussion with @jits, we want to give pnpm a try.

---

I find pnpm _smoother_ (gut feel), after many years of using yarn (and I've switched over to pnpm for all Node.js based projects now).

A few concrete reasons why this might be the case:

- **performance** — because pnpm uses symlinks to point your dependencies to a single source of truth on disk, it's very fast. 
- **better dependency isolation** — pnpm doesn't flatten the dependencies in the `node_modules` folder, so (in theory) provides a safer isolation between conflicting versions of sub-dependencies. It will still report conflicting versions though (so you can take action). My experience has been smoother in these situations with pnpm than yarn, where the latter would decide for you which version of the sub-dependency should be installed.
- **support** — Yarn v1 (which is what `npx yarn` would fetch and use) has _kinda_ been deprecated / is unsupported (https://github.com/yarnpkg/yarn?tab=readme-ov-file#%E2%84%B9%EF%B8%8F-important-note). It's not clear to me how long of a future it has. And the compatibility and migration story between v1 and subsequent versions isn't particularly smooth (https://yarnpkg.com/migration/guide). I've had annoying experiences migrating from Yarn v1 to a later version.
  - Whereas pnpm is gaining a lot of traction — e.g. it's the preferred choice in the Nuxt and unjs ecosystem.

I think I had a similar experience to you a few years back, but now it seems pnpm has matured quite a bit.

(https://refine.dev/blog/pnpm-vs-npm-and-yarn/ is a more detailed comparison, if you want to dig further)

_Originally posted by @jits in https://github.com/angular-architects/ngrx-toolkit/issues/72#issuecomment-2250046837_

jits:
Closes #73

I've verified that:

- Running `pnpm start` works locally
- Running `pnpm build:all` works locally
- CI passes (using pnpm exclusively)
- Caching works as expected in subsequent CI builds

Note: running `pnpm verify:all` causes some lint errors and a test error due to no test suite(s).

---

TODO:

- [x] Ensure CI passes
- [x] Update README(s?) to mention pnpm is used for local development


.github/workflows/build.yml
```diff
@@ -8,14 +8,17 @@ on:
       - main
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
jits, line 11: @rainerhahnekamp — I feel it's better to lock down to specific versions wherever possible, hence this change. But am happy to revert this bit if you feel otherwise.

jits:
@rainerhahnekamp — for the README update (to mention pnpm is used for local development), I'm thinking of quickly mentioning it under the "I have an idea for a new extension, can I contribute?" FAQ entry. Does that work for you, for now? In the long run it may be worth having a dedicated CONTRIBUTING doc.

rainerhahnekamp:
Thanks, and about the README, yeah sure, why not?

jits:
@rainerhahnekamp — ready for your review (when you have a sec) :)

package.json
```diff
@@ -3,11 +3,15 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
-    "start": "npx nx serve demo",
-    "build:all": "npx nx run-many --targets=build",
-    "verify:all": "npx nx run-many --targets=lint,test,build"
+    "start": "nx serve demo",
```
rainerhahnekamp, line 6: why did you remove `nx`?

.github/workflows/build.yml
rainerhahnekamp: Nope, that's perfectly fine

rainerhahnekamp (commented):
Left a comment about the removal of nx

package.json
jits: I take it you mean the removal of `npx`? It's no longer needed as `npx` won't work with pnpm (I think), and is not necessary (I believe).

rainerhahnekamp:
Sorry, I will never understand this. I selected "rebase and merge" and GitHub states me as the committer. If you want I can rebase main to the former, you can create the same PR again, and I try a simple merge.

jits:
@rainerhahnekamp — that's not a problem at all — it shows as "jits authored and rainerhahnekamp committed" :)

rainerhahnekamp:
Alright, then thanks a lot again.

PR #84, opened 2024-08-11
docs: Add missing `updateAll` example to `withDataService` interface examples (root & lib)

michael-small:
I remembered seeing `withDataService` after I read an article that did something like it. I was going to tweak this to use `rxMethod` for my own project, but then I realized the doc pages were missing `updateAll`.


rainerhahnekamp (approved):
LGTM

rainerhahnekamp:
Thanks Michael

Issue #85, opened 2024-08-11
suggestion/question: `withDataService` support for Observables


michael-small:
## My problem

[`withDataService`](https://github.com/angular-architects/ngrx-toolkit/tree/42d098b2140f63692085ce949c88b4d8250e8870?tab=readme-ov-file#dataservice-withdataservice) is super smart and interesting. Honestly the most boilerplate reducing Angular code I have seen. One limitation for my own use, however, is that I use `rxMethod` rather than `promise`.

## My suggestions/questions

I copied over the relevant files of this library's `withDataService` into a private project full of more signal store experimentation I adapted all the feature methods to use `rxMethod`. I have not tested it too much yet, but it seems to work good so far. I can make it public if anyone is interested.

1. If feasible, would there be interest in expanding `withDataService` to support RXJS? 
    - I have an implementation working, but not yet with equivalent tests and not tested manually that much. If there is interest I would handle that, as well as documentation.
    - I certainly couldn't have written this from the ground up - I imagine that doing this would add a lot to the complexity. 
2. If there is not interest, how would I attribute this library properly so that I can use my variation under the MIT License
    - I may be overthinking things, but I haven't done something like this before: how should I go about fulfilling the whole MIT clause "The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software."?
    - Include a markdown of it in the root of the feature?

edit: link to my implementation in one of my projects: https://github.com/michael-small/signal-store-playground/tree/main/src/app/store/withDataService-rxjs

rainerhahnekamp:
@michael-small, I don't see a problem in supporting RxJs. So feel free to contribute your implementation

michael-small:
Sounds good. I in fact do have some changes to make as I messed some stuff up, but once I have it in a better state I will update this issue and then get to a PR.

michael-small:
Link to my implementation in a project: https://github.com/michael-small/signal-store-playground/tree/main/src/app/store/withDataService-rxjs

Once I have ironed it out a bit more I will make a fork of the toolkit and start adding an RXJS implementation.

michael-small:
I have been grappling with types and generics when trying to adapt `withDataService` to also accept Observables. After trying and throwing away a lot of approaches, I think I found a decent one, and I was wondering if this sounds like a decent approach. The generic names are quite rough.

```ts
export interface DataService<
  E extends Entity,
  F extends Filter,
[...]

PR #88, opened 2024-08-27
feat: add support for observables to `withDataService`

michael-small:
This PR would close issue 85: [suggestion/question: withDataService support for Observables](https://github.com/angular-architects/ngrx-toolkit/issues/85)

What was done
- The interface for the implementing services that use `withDataService` was expanded to take Observable methods
    - `type PromiseOrObservable<Entity> = Promise<Entity> | Observable<Entity>;`
    - ex) `loadById(id: EntityId): PromiseOrObservable<Entity>;`
        - Existing functionality supports something like this
            -   `loadById(id: EntityId): Promise<Flight> {...}`
        - These changes would now also support something like this
            -  `loadById(id: EntityId): Observable<Flight> {...}`
- The `withDataService` methods were adapted to take an observable (returns a subscription) as well as already taking promises
    - `type PromiseOrSubscription<Entity> = Promise<Entity> | Subscription;`
    - ex) ```[K in Collection as `load${Capitalize<K>}Entities`]: () => PromiseOrSubscription<void>;```
        - The method checks if something is an instance of an observable. If so, the method manually pipes and subscribes. If not, the method resolves the call as a promise. 

edit:
What was chosen not to be done, as per talk in the issue
- For the moment, `rxMethod` is not used. The solution currently just manually subscribes. Supporting `rxMethod` would widen the feature to need to take non-static values such as signals or observables. 
    - "I'd say for the first version, it should be fine of you check if the response is of type Promise or Observable. I guess you would do an automatic subscribe. If there's demand, we can later maybe add the rxMethod and map a Promise to an Observable." - [Rainer](https://github.com/angular-architects/ngrx-toolkit/issues/85#issuecomment-2308991952)

What is left to be done (before turning from Draft PR to regular PR)
- Tests for both the promises and observables
    - There are not already existing tests for `withDataService` as is with promises, so I should probably make those first and get those in as a PR before moving onto this one. Thoughts?
        - edit: See the next PR: https://github.com/angular-architects/ngrx-toolkit/pull/89
    - Observable tests should follow
    - I am admittedly not very experienced/good at testing, so testing is probably the main hurdle now that implementation is hopefully all good apart from iteration on feedback.
- Explicitly handle unsubscribing
    - HTTP observables tend to be considered self handling observables if I understand correctly, but it would be good to unsubscribe anyways
    - Pass in an optional `DestroyRef`? Otherwise basic `take(1)` or `first` or something?


michael-small:
I forgot to actually push up the usage of the RXJS equivalent service and point that service at the right endpoint... fixed https://github.com/angular-architects/ngrx-toolkit/pull/88/commits/47fc853d373905dd81c267ecc8ac972e071350f8

michael-small:
Closing for now out of interest + time tbh

jryan719:
@michael-small  @rainerhahnekamp , any chance this will get picked up again to be officially supported by the ngrx-toolkit library? I have a huge need for it, but didn't want to introduce it to my project if the spec was likely to change after official adoption... Thoughts?

Issue #53, opened 2024-05-31
Possibility to exclude withDevTools in production mode when ngrx-toolkit is installed under dependencies


koempf:

REVenkatesan:
Is there any possibility to exclude withDevtools in production mode...

marcindz88:
Hi there I have a solution for this although it is a bit hacky: 
Assuming you have setup environment files you can do the following: 

environment.ts:
```typescript
import { withDevtools } from '@angular-architects/ngrx-toolkit';

export const environment = {
   storeDevToolsFeature: withDevtools
}
```

environment.prod.ts
```typescript
export const environment = {
   storeDevToolsFeature: (_: string) => store => store
}
```

If you have a type defined for your environment, you can use the type: 
```typescript
{
  storeDevToolsFeature: (name: string) => SignalStoreFeature
}
```

Then all you need to do is replace `withDevTools` everywhere in your app with `environment.storeDevToolsFeature`
e.g.: 
```typescript
export const SomeStore = signalStore(
  withState({strings: []. as string[] }),
  environment.storeDevToolsFeature('some')
);
```

Make sure you have file replacements in angular.json prod configuration:
```json
"fileReplacements": [
  {
    "replace": "src/environments/environment.ts",
    "with": "src/environments/environment.prod.ts"
  }
]
```

As a result, when running the prod version of ng serve or build this file will be replaced and not compiled at all, which will disable and tree-shake the whole withDevTools feature 😀

rainerhahnekamp:
We could use `isDevMode` (https://angular.dev/api/core/isDevMode) instead of accessing the environment files. As far as I understand, that's the standard way on how to do it.

`withDevtools` could become a wrapper function which first checks if the user wants to disable it during prod mode and could then return an empty Feature.

@marcindz88 would you have time to implement that and provide a PR?

marcindz88:
@rainerhahnekamp I thought about adding isDevMode but this leads to the withDevTools code to be included in the production bundle, unless there is some method to mark it for exclusion that I don't know.

marcindz88:
Hmm, I may be wrong about that, I will check if it is tree shaken by default and if yes, I will provide a PR

REVenkatesan:
i have tried as you post @marcindz88 but i got the below error while importing environment file of production
export const environment = {
  storeDevToolsFeature: (_: string) => (store: any) => store
};
Spread types may only be created from object types.ts(2698)
[...]

PR #94, opened 2024-09-14
feat: withDevTools disabled in prod - tree-shaking docs

marcindz88:
Solution to #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,9 @@ updateState(this.store 'update loading', {loading: false});
 
   <summary>Devtools tree-shaking details</summary>
 
-  environment.ts:
+It is required to add the `withDevtools` function to the environment files.
+
+environments/environment.ts:
 ```typescript
 import { withDevtools } from '@angular-architects/ngrx-toolkit';
 
@@ -91,7 +93,7 @@ export const environment = {
 }
 ```
 
-environment.prod.ts
+environments/environment.prod.ts
 ```typescript
 import { withDevtoolsStub } from '@angular-architects/ngrx-toolkit';
 
@@ -100,12 +102,21 @@ export const environment = {
 }
 ```
 
-Then all you need to do is replace `withDevTools` everywhere in your app with `environment.storeWithDevTools`
+Then you can create utility function which can be used across the application
 e.g.:
+
+shared/store.features.ts (or any other file)
+```typescript
+import { environment } from 'src/environments/environment';
+
+export const withTreeShakableDevTools = environment.storeWithDevTools;
+```
+
+And use it in your store definitions:
 ```typescript
 export const SomeStore = signalStore(
   withState({strings: [] as string[] }),
-  environment.storeWithDevTools('featureName')
+  withTreeShakableDevTools('featureName')
 );
 ```
 
--- a/README.md
+++ b/README.md
@@ -76,6 +76,52 @@ patchState(this.store, {loading: false});
 updateState(this.store 'update loading', {loading: false});
 ```
 
+`withDevtools()` is by default enabled in production mode, if you want to tree-shake it from the application bundle you need to abstract it in your environment file.
+
+<details>
+
+  <summary>Devtools tree-shaking details</summary>
+
+  environment.ts:
+```typescript
+import { withDevtools } from '@angular-architects/ngrx-toolkit';
+
+export const environment = {
+  storeWithDevTools: withDevtools
+}
+```
+
+environment.prod.ts
+```typescript
+import { withDevtoolsStub } from '@angular-architects/ngrx-toolkit';
+
+export const environment = {
+  storeWithDevTools: withDevToolsStub
+}
+```
+
+Then all you need to do is replace `withDevTools` everywhere in your app with `environment.storeWithDevTools`
+e.g.:
+```typescript
+export const SomeStore = signalStore(
+  withState({strings: [] as string[] }),
+  environment.storeWithDevTools('featureName')
+);
+```
+
+Also make sure you have defined file replacements in angular.json prod configuration:
+```json
+"fileReplacements": [
+  {
+    "replace": "src/environments/environment.ts",
+    "with": "src/environments/environment.prod.ts"
+  }
+]
+```
+
+</details>
+
+
 ## Redux: `withRedux()`
 
 `withRedux()` bring back the Redux pattern into the Signal Store.
--- a/libs/ngrx-toolkit/src/index.ts
+++ b/libs/ngrx-toolkit/src/index.ts
@@ -1,4 +1,5 @@
 export {
+  withDevToolsStub,
   withDevtools,
   patchState,
   updateState,
--- a/libs/ngrx-toolkit/src/lib/with-devtools.ts
+++ b/libs/ngrx-toolkit/src/lib/with-devtools.ts
@@ -80,6 +80,11 @@ export function reset() {
 }
 
 /**
+ * Stub for DevTools integration. Can be used to disable DevTools in production.
+ */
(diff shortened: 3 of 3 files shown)


rainerhahnekamp:
Thanks @marcindz88. In the meantime, I've looked up the approach of the NgRx Devtools.

The approach for tree-shaking is very similar to what you proposed, namely, using the environment files.

Disabling it with `isDevMode()` does not exist. Instead, the user can set a `logOnly` property, which disables certain features like time traveling.

What if we provide `logOnly` as an option that is enabled by default instead? 

Time traveling is unavailable at the moment, so it would not have an immediate effect.

Users could use the extension like 

`withDevtools('flights': options: {logOnly: !isDevMode()})`?

And the signature could be 

```typescript
type Options = {
  logOnly: boolean
}
withDevtools(name: string, options: Options = {logOnly: false);
```

marcindz88:
@rainerhahnekamp  Yeah I was thinking, that some users may want to enable devtools in some prod-optimized environments like test staging. However I don't really like repeating the configuration in every store in app. 

Does this feature work in injection context? If yes maybe we could provide a provider that creates an injection token with global ngrx-toolkit configuration?

rainerhahnekamp:
@marcindz88 

> However I don't really like repeating the configuration in every store in app.

You are right, I am all for an "injection" token approach. This would also fix the issue with one store having `logOnly` set to to `true` and another to `false`.

`withDevtools` has access to the injection context and as soon as it becomes relevant, i.e. it supports time travel, it can request the value of the injection token.

---

Btw, why do you disable dev tools in a production environment? I always found it extremely helpful for debugging and reproducing issues that only occur in production.

It is not that we are exposing any secrets. Anyone can monitor the network requests and—with a little bit of work—find out what's going on by looking at the minimized Javascript files.

Just curious to learn something new here.

marcindz88:
I think there are several arguments for disabling / removing it in production:

1. Performance - Redux monitoring tool can become an overhead for application, as it stores the state and history of states of the application. I have seen it lag in large stores.
2. Bundle size
3. While a user can always find everything in obfuscated javascript files, it requires very good knowledge of Angular and is still non-trivial to observe actions and the flow of application state. It is hard to find and see the state of application without using debugger and knowledge about the app structure. With the Redux extension everything is visible for anyone, who has installed it, without much effort.


Do you think we should add one configuration token for the whole library or a per-feature approach?

rainerhahnekamp:
Thanks for the detailed explanation. I suggest we go with one global configuration token. It should be just on or off

marcindz88:
@rainerhahnekamp please take a look. 

I have observed while testing that the store does not disconnect after application restart, I think this is because connect observer is not unsubscribed nor is disconnect called anywhere FYI https://github.com/reduxjs/redux-devtools/blob/main/extension/docs/API/Methods.md#disconnect. 

Also, I was thinking and we could use the pattern of overriding action creators to be able to detect action dispatches without the use of `updateState`.

marcindz88:
@rainerhahnekamp please also take a look at this experiment: https://github.com/angular-architects/ngrx-toolkit/commit/1c7ee5d597724b08d2f879f9dd451bdf83658737
It simplifies the code a bit and now whole actions with payload are visible in redux dev tools with no need for overriding of patchState. 
However there is an issue with private actions - because these are not methods in the store they are not visible, but we could probably work around this somehow. Also the new dispatch action approach is probably not compatible, but we could override dispatch fn with the dev tools logic.

rainerhahnekamp:
@marcindz88 thanks, I have another branch that deals with disconnection issues and some other stuff but haven't found the time to push it yet.

I'll also checkout the automatic action name detection change, although I still think `updatePatch` has its use case when developers want to provide a specific, readable name for the action.

libs/ngrx-toolkit/src/lib/with-devtools/with-devtools.ts
```diff
@@ -87,7 +88,8 @@ export function withDevtools<Input extends EmptyFeatureResult>(
 ): SignalStoreFeature<Input, EmptyFeatureResult> {
   return (store) => {
     const isServer = isPlatformServer(inject(PLATFORM_ID));
-    if (isServer) {
+    const { logOnly } = inject(DEVTOOLS_CONFIG, { optional: true }) || DEFAULT_DEVTOOLS_CONFIG;
+    if (isServer || logOnly) {
```
rainerhahnekamp, line 92: `logOnly` doesn't have an effect at the moment. Logging still happens but features like time travelling will be disabled.

Since we "only" have logging at the moment, we can leave this condition with `isServer` only.

libs/ngrx-toolkit/src/lib/with-devtools/with-devtools.config.ts
```diff
@@ -0,0 +1,35 @@
...
+ * ```ts
+ * provideStoreDevtoolsConfig({
+ *  logOnly: !isDevMode(), // Enable the log-only mode when not in production. (Default: false)
+ * });
+ *  ```
+ *
+ *  @returns The provider of the custom configuration.
+ */
+export const provideStoreDevtoolsConfig = (config: Partial<DevtoolsConfig>) => {
+  return {
```
rainerhahnekamp, line 28: Do we want to make that as an environment provider so that it could only be called in the root scope or in paths?

What about override protection? I think that would also be useful.

rainerhahnekamp (commented):
Good work so far

libs/ngrx-toolkit/src/lib/with-devtools/with-devtools.ts
marcindz88: I don't know if we should include this config for now. Maybe let's stash it for later (when it will do something) and merge only the tree-shaking docs ?

marcindz88:
@rainerhahnekamp regarding my experiment version: It is not only about the actionName, we should also pass action payload to redux extension. Also, we could prefix the action name with [name of store] and possibly change camelcase to a more readable version.

libs/ngrx-toolkit/src/lib/with-devtools/with-devtools.ts
rainerhahnekamp: I am fine with merging only the docs

rainerhahnekamp:
@marcindz88 In order to get only the documentation part merged, could you please remove all files except the README from your PR?

marcindz88:
@rainerhahnekamp PR updated, I also added an a stub const for easier use in the environment file. Let me know if that's ok.

README.md
```diff
@@ -76,6 +76,52 @@ patchState(this.store, {loading: false});
...
+environment.prod.ts
+```typescript
+import { withDevtoolsStub } from '@angular-architects/ngrx-toolkit';
+
+export const environment = {
+  storeWithDevTools: withDevToolsStub
+}
+```
+
+Then all you need to do is replace `withDevTools` everywhere in your app with `environment.storeWithDevTools`
```
rainerhahnekamp, line 103: Do you think it is possible to update the example with a utility function which encapsulates the access to environment?

I don't think it is a good idea if we promote spreading access to the environment in the codebase.

So something like 

```typescript
function withTreeShakableDevTools(featureName: string) {
  return environment.storeWithDevTolls(featureName);
}
```

rainerhahnekamp (changes requested):
Wonderful, I just left a small comment on a stylistic issue but then we are ready to go.

README.md
marcindz88: I added an exported const which lets us not to import environment file directly, and allows for keeping the function signature internal to the library (may be useful if there were changes)

rainerhahnekamp (approved):

rainerhahnekamp:
Thanks again @marcindz88

PR #102, opened 2024-11-04
chore: adjusted version range to allow major version of @ngrx/signal …

### LINKED ISSUE #101: Increase possible version range of @ngrx/signals
Currently the version is pinned to exactly 18.1.0. Maybe change the version range to ^18.1.0 or at least ~18.1.0.

thanks for your work! 👍

Zwergenpunk:
closes #101 

I removed the readme part, but I'm not sure what you want. 
A table that just says version 18 is ok with 18 does not make any sense. Maybe you can add it back with version 19? 🤷 
or we just make it 18/18 so that the skeleton is ready for new versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,31 +19,13 @@ To install it, run
 npm i @angular-architects/ngrx-toolkit
 ```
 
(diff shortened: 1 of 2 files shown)


rainerhahnekamp (approved):
Nice

rainerhahnekamp:
Thanks @Zwergenpunk

PR #106, opened 2024-12-09
feat: Added support for clear undo redo stack.

LukasSefcik:
Added function clearing both undo a redo stack.
New test verifying undo redo functionality.
Updated README.md.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
@@ -192,32 +192,23 @@ describe('withUndoRedo', () => {
       });
     }));
 
-    it('clears undo redo stack', fakeAsync(() => {
-      TestBed.runInInjectionContext(() => {
-        const Store = signalStore(
-          withState(testState),
-          withMethods(store => ({ update: (value: string) => patchState(store, { test: value }) })),
-          withUndoRedo({ keys: testKeys })
-        );
+    it('clears undo redo stack', () => {
+      const Store = signalStore(
+        { providedIn: 'root' },
+        withState(testState),
+        withMethods(store => ({ update: (value: string) => patchState(store, { test: value }) })),
+        withUndoRedo({ keys: testKeys })
+      );
 
-        const store = new Store();
-        tick(1);
+      const store = TestBed.inject(Store);
 
-        store.update('Foo');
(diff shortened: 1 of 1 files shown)


libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
```diff
@@ -190,5 +191,33 @@ describe('withUndoRedo', () => {
         expect(store.canRedo()).toBe(true);
       });
     }));
+
+    it('clears undo redo stack', fakeAsync(() => {
+      TestBed.runInInjectionContext(() => {
```
rainerhahnekamp, line 196: I know that you did it as the other tests before, but could please do it without runinInjectionContext? So you setup the signalStore first (with providedIn root), and then you just say TestBed.inject(Store)?

libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
```diff
@@ -190,5 +191,33 @@ describe('withUndoRedo', () => {
         expect(store.canRedo()).toBe(true);
       });
     }));
+
+    it('clears undo redo stack', fakeAsync(() => {
```
rainerhahnekamp, line 195: `fakeAsync` is going to be not available in zoneless and it is not required in this test. If you remove the `tick`, the test should still work.

Out of interest, why do you see the need for a `tick`?

rainerhahnekamp (changes requested):
@LukasSefcik well done, two remarks and then we are ready to merge.

libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
LukasSefcik: You are absolutely right. No need to have `fakeAsync` and `tick`. I used `tick` automatically based on my previous experiences with some not working tests in my projects. Sorry for this.

libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
LukasSefcik: It's fixed. Thank you

rainerhahnekamp (approved):

rainerhahnekamp:
Thanks @LukasSefcik

PR #120, opened 2025-01-07
docs: add missed words in `withReset()` description

michael-small:
I was checking out the new docs, they are nice to have. Noticed this phrasing was missing a couple words.

Also, `withReset()` is not listed in the sidenav as it is deployed now by the way.


rainerhahnekamp:
Thanks @michael-small

Issue #111, opened 2024-12-30
withStorageSync support for INDEXEDDB


audhit:
withStorageSync is working and a cool feature, thanks a lot for it, but I would love to see the support for **indexeddb** as well because it has more storage limitation compare to 4 MB of localstorage. Please let me know, how can I achieve it. I really want the storage to be **indexeddb** instead of localstorage or sessionstorage. Thanks in advance.

rainerhahnekamp:
You can add a PR. If you are quick, it might land in the 19 release (which is overdue)

audhit:
thanks @rainerhahnekamp , is there any 3rd party tool which can provide storage persistence with indexeddb? Please let me know, thanks :)

rainerhahnekamp:
@audhit I only know https://dexie.org/

audhit:
@rainerhahnekamp do we now have the support for indexedDB storage instead of localStorage/sessionStorage in the latest version v19.0.2 or it is still not there? Please let us know, thanks in advance for the great library...keep rocking 🚀

rainerhahnekamp:
Not yet, but nobody prevents us from adding it. I was waiting for further input from the community about the recommended tooling. I don't have a problem with dexie but would know if it is still the library to use.

audhit:
https://dexie.org/ looks really good, it is possible to create a wrapper out of dexie and connect it with storageSync API just like **below** to make it as simple as possible, hoping for the best

```const SyncStore = signalStore(
  withStorageSync<User>({
    storage: () => dexie,  // rest of the things will be taken care automatically by the library without much extra configs maybe 
  })
);`

mzkmnk:
Would @audhit  like to add a PR?

rainerhahnekamp:
@audhit or @mzkmnk or both: You can give it a try. The only design requirement that I have, is that it needs to be treeshakable. We only want to have dexie as requirement if the user wants to use it.

Therefore:

- The syntax should be `withStorageSync({}, withDexie())` (as it is in `withDevtools`)
- `withDexie` needs to be a secondary entrypoint (look at the redux connector as example)

mzkmnk:
@rainerhahnekamp 
[...]

PR #134, opened 2025-02-02
feat(with-storage-sync): indexeddb support

mzkmnk:
### Issue
#111 

### Content

In this PR, we modified `withStorageSync` to support IndexedDB.

### Breaking Changes
- The storage argument has been removed, and a new second argument now allows you to choose `withIndexeddb()`, `withLocalStorage()`, or `withSessionStorage()`. By default, `withLocalStorage()` is used.

### Sample Code
```ts
const TodoIndexedDBSync = signalStore(
  withState<...>({...}),
  withStorageSync(
    {key:'todos'},
    withIndexeddb(),
  ),
  withMethods((store) => ({...}))
);
```


mzkmnk:
@rainerhahnekamp
The code is complete. Please review it! :octocat:

mzkmnk:
@audhit
What do you think of this fix？

audhit:
> @audhit What do you think of this fix？

Hey @mzkmnk looks promising. This is exactly what I am looking for. NGRX tool will be complete with this feature to be honest. @rainerhahnekamp please verify and merge it soon if possible and let us know about the new release. Thanks a lot for the great work guys 🙏

libs/ngrx-toolkit/src/lib/storage-sync/tests/with-storage-sync.spec.ts
```diff
@@ -14,16 +16,21 @@ const initialState: StateObject = {
 const key = 'FooBar';
 
 describe('withStorageSync', () => {
```
rainerhahnekamp, line 18: Is it possible that we run all tests twice? One with the default and one with indexeddb adapter? So you would wrap a for loop over the tests where the first run uses the default and the second the indexeddb adapter?

I think we don't have then to come up with separate tests for indexeddb itself.

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
```diff
@@ -0,0 +1,220 @@
...
+  key: string;
+  /**
+   * Allows selection between localStorage, sessionStorage, and indexedDB
+   *
+   * Defaults to `localStorage`
+   */
+  storageType?: 'localStorage' | 'sessionStorage';
+};
+
+export type IndexedDBSyncConfig<State> = BaseSyncConfig<State> & {
```
rainerhahnekamp, line 80: from a design perspective, it would be better , if a user just has to switch from local to index or the other way around without adding any other additional parameters.

localStorage is always bound to the domain, which means it kind of uses the domain as `dbName` and the key as `storeName`.

Would it be possible, that we just stick to `key` and `dbName` would automatically be the domain name and `storeName` would be `key`?

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
```diff
@@ -0,0 +1,220 @@
...
+};
+
+export type Config<State> = SyncConfig<State> | IndexedDBSyncConfig<State>;
+
+/**
+ * Enables store synchronization with storage.
+ *
+ * Only works on browser platform.
+ */
+export function withStorageSync<Input extends SignalStoreFeatureResult>(
```
rainerhahnekamp, line 105: I know it is tricky, but I meant that indexedDB is activated via a `with` method.

So if users would run `withStorageSync('flights')`, we would use localStorage. If they would run `withStorageSync('flights', withIndexedDB())`, we would switch to indexDB.

We need to provide the function `withIndexedDB()`, and it could return the `IndexedDBService` class. If `withIndexedDB()` is not provided, then it would use a `withLocalStorage()` as default.

Both `withIndexedDB()` and `withLocalStorage()` would return the Service classes.

So when I can call `withIndexedDB()`, I'd get `IndexedDBService`, otherwise `LocalStorageService` (which does not exist yet).

Since we would then have two different implementations of a "StorageService", we would also come up with an interface that both `withLocalStorage` and `withIndexedDB` return. The name of it could be `StorageService` and would replace the existing class.

In short, the implementation could look something like this:
```typescript
type StorageServiceFactory = () => Type<StorageService>;

const withIndexedDB: StorageServiceFactory = () => IndexedDBStorageService;
const withLocalStorage: StorageServiceFactory = () => LocalStorageService;

export function withStorageSync<
  State extends object,
  Input extends SignalStoreFeatureResult
>(
  configOrKey: string | Config<State>,
  StorageServiceClass = withLocalStorag() // <-- this was added
): SignalStoreFeature<Input, WithStorageSyncFeatureResult> {

  // initialization logic...

  return signalStoreFeature(
    withMethods(
      (
        store,
        platformId = inject(PLATFORM_ID),
        storageService = inject(StorageServiceClass) // <-- this changed
      ) => {
```

rainerhahnekamp (changes requested):
Very good work!

I think my biggest suggestion would be this refactoring to `withIndexedDB` and `withLocalStorage()`.

Other than that, please make sure that the Git History is kept intact (I've added a comment about it).

If there is something unclear, please let me know. I have now a good overview and should be able to respond promptly.

rainerhahnekamp:
@mzkmnk an additional explanation, via it should be `withIndexedDB()` and not it is right now.

With the current implementation, we have the `IndexedDB` service in the bundle, even if it is not used. If that Service only is imported via `withIndexedDB()` and no one is calling/importing `withIndexedDB()`, it never ends up in the bundle.

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
mzkmnk: I decided to use keys for both storeName and dbName. The reason is that if we used only storeName as a key, we’d need to manage the database version, which would deviate from our main goal this time.

libs/ngrx-toolkit/src/lib/storage-sync/tests/with-storage-sync.spec.ts
mzkmnk: For the time being, I’ve modified it so that the first run uses localStorage and the second run uses IndexedDB. Please check if this works.

mzkmnk:
@rainerhahnekamp 
I’ve finished making the changes. Please review them!

libs/ngrx-toolkit/src/lib/with-storage-sync.ts
```diff
@@ -75,71 +77,96 @@ export type SyncConfig<State> = {
...
+      ) => {
+        if (isPlatformServer(platformId)) {
+          console.warn(
+            `'withStorageSync' provides non-functional implementation due to server-side execution`
+          );
+          return StorageSyncStub;
+        }
 
-      const storage = storageFactory();
+        return {
```
rainerhahnekamp, line 124: Sounds a little bit crazy, but since we are delegating all the logic to the StorageService, wouldn't it be possible to make `withStorageSync` an overloaded function where `withLocalStorage` is non-Promised based and `withIndexedDb` Promise?

In that regard, we wouldn't introduce breaking changes and keep up the DX for local storage.

rainerhahnekamp (changes requested):

libs/ngrx-toolkit/src/lib/with-storage-sync.ts
```diff
@@ -75,71 +77,96 @@ export type SyncConfig<State> = {
...
-        return StorageSyncStub;
-      }
+    withMethods(
+      (
+        store,
+        platformId = inject(PLATFORM_ID),
+        storageService = inject(StorageServiceClass)
+      ) => {
+        if (isPlatformServer(platformId)) {
+          console.warn(
```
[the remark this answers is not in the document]
mzkmnk: fixed here
[f1e9c8b](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/f1e9c8bbd84d739dcf843b6dae933220effcc5e3)

audhit:
@rainerhahnekamp looks good to me, can you please kindly test @mzkmnk's solution and let us know if you can merge and release it? It would be a great help in my project. Thanks in advance 🙏

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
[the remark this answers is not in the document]
rainerhahnekamp: Uhm, sorry, but it looks like Git didn't see it as moving the file but sees it is a whole new file.

rainerhahnekamp:
@mzkmnk 

> using the IndexedDB keyPath as the key, multiple pieces of data can be stored in a single database

So we are saying we are using the same `dbName` and `dbName` for all instances, right? If yes, please implement that 👍

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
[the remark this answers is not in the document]
mzkmnk: @rainerhahnekamp 
It was detected as a new file even when using git mv command.

mzkmnk:
@rainerhahnekamp 

>So we are saying we are using the same dbName and dbName for all instances,

This one has already been corrected.Please check.

[55e81df](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/55e81df9c07dc3ee2bc262ec39878f0245c5ac14)

libs/ngrx-toolkit/src/lib/with-storage-sync.ts
mzkmnk: @rainerhahnekamp 
I can’t come up with a good way to implement this. I tried using overloaded functions for now (here is the relevant commit: [c041a49](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/c041a4968d7827e44106b081291aa53ce50905f8)).

When I use `withLocalStorage()` (or `withSessionStorage()`), the returned function itself is not a Promise, but it’s using a Promise internally. I suspect the implementation isn’t ideal. Do you have any good solutions or ideas for this?

audhit:
> @audhit, we are in the middle of the second set of change requests. We can't merge now.

Waiting for it eagerly 🚀 , please kindly make it happen soon if possible 🚀  @rainerhahnekamp .... thanks a lot in advance

libs/ngrx-toolkit/src/lib/storage-sync/with-storage-sync.ts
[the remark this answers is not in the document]
rainerhahnekamp: Sorry to bother you with this. We can make a compromise. You could move it back (outside of storage-sync) so that the Git History is not destroyed. We will merge and then maybe do a refactoring where we just do the git move itself.

libs/ngrx-toolkit/src/lib/with-storage-sync.ts
rainerhahnekamp: I'll leave it for later then.

rainerhahnekamp (changes requested):
@mzkmnk I think we should get that issue with git history sorted out and then merge it.

As I said in a comment before. Maybe it is better to move the with-storage-sync.(spec).ts files back where they were and we'll try to do the "git mv" in a separate branch.

mzkmnk:
@rainerhahnekamp 

I understand. I will move with-storage-sync.(spec).ts back to its original location. Also, regarding [c041a49](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/c041a4968d7827e44106b081291aa53ce50905f8), since it's not really a good fix, would it be okay to revert the changes?

rainerhahnekamp:
@mzkmnk 
> I understand. I will move with-storage-sync.(spec).ts back to its original location. Also, regarding [c041a49](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/c041a4968d7827e44106b081291aa53ce50905f8), since it's not really a good fix, would it be okay to revert the changes?

I would leave everything as it is, so no reverting. The only thing is really about keeping the git history intact. That should do it.

mzkmnk:
@rainerhahnekamp 

I've completed moving the with-storage-sync.(spec).ts file.
[c45b665](https://github.com/angular-architects/ngrx-toolkit/pull/134/commits/c45b665c5fc023371224828b11b148d75fea83c2)

rainerhahnekamp:
Wonderful @mzkmnk. I'm going to squash and merge it. Congratulations!

rainerhahnekamp:
@mzkmnk do you want to create the PR which just moves with-storage-sync.*.ts files or should I do that?

mzkmnk:
@rainerhahnekamp 

After this PR is merged, I’ll create a PR to move the with-storage-sync.(spec).ts files.

audhit:
wowowow...great work guys...now new release and documentation update needs to be done...its soo great and good feeling when someone's request is taken care of. I am so thankful to you guys @mzkmnk  and @rainerhahnekamp ... Thank you so much, waiting for it to be released...thanks again :)

rainerhahnekamp:
Full kudos ot @mzkmnk

mzkmnk:
@rainerhahnekamp 
@audhit 
My first OSS development experience was a great memory. Thank you!

PR #140, opened 2025-02-09
docs: make snippet langs consistent & fix example indent

michael-small:
I looked at the docs snippet after noticing the last PR and noticed two things

- One snippet needed indents (`withDataService`)
- Some snippets didn't have a language, or were not consistent with others


rainerhahnekamp:
Thanks, yeah according to ChatGPT using `typescript` instead of `ts` is better!

PR #141, opened 2025-02-11
doc: Storage factory explanation on withStorageSync()

DJREMiX6:
I added an example that explains how you can use the storage factory to inject a service to be used as a different storage system rather than the LocalStorage or SessionStorage.

The use case was that i wanted to sync the state in a way that was not reachable and updateable by the user so i created a Storage implementation which holds everything in memory, every time you destroy a component in which the state is provided and you recreate it, the state is synced with that in memory Storage and the user will not be able to read it or change it.

### DIFF
--- a/docs/docs/with-storage-sync.md
+++ b/docs/docs/with-storage-sync.md
@@ -41,3 +41,39 @@ public class SyncedStoreComponent {
   }
 }
 ```
+## Injecting services
+The `storage` property requires a factory function that returns a `Storage` implementation. If you have a custom `Storage` implementation that is `Injectable`, you can provide it as a factory parameters using the `inject` function.
+
+ Example:
+ 
+ ```typescript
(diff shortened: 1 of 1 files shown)


rainerhahnekamp:
Is it possible to merge that after #134?

DJREMiX6:
> Is it possible to merge that after #134?

Sure is! Should I wait for it to be merged and rebase or will you do it?

rainerhahnekamp:
A quick update here. We have in the meantime the indexeddb support, but I'd like to modify it so that we have minimal changes, i.e. get rid of async from synchronous stores (local & session storage).

rainerhahnekamp:
@DJREMiX6, has been a while but the indexeddb support is now available. The documentation has also been updated. If you think your PR could improve the docs, please re-open it. Closing it for now.

Issue #147, opened 2025-02-16
Feature Request: Enhance withCallState to Support Multiple Collections


abhijit-chikane:
**Feature Request: Enhance withCallState from @angular-architects/ngrx-toolkit to Support Multiple Collections and make a part of core @ngrx/signals package**

**Issue**
Currently, withCallState only accepts a collection with a single string value. However, in large projects, we often need to track the loading state of multiple collections simultaneously.

While it's possible to use multiple withCallState functions, this approach has a significant drawback: the Store currently accepts only 15 inputs. Using withCallState multiple times can quickly reach this limit, making state management inefficient.

**Proposed Solution**
Extend withCallState to accept an array of strings, allowing developers to manage the state of multiple collections more effectively.

I've already implemented a similar feature in a custom store and shared my approach in this [Medium article](https://medium.com/@abhijit_chikane/simplifying-request-status-management-in-angular-with-ngrx-signal-store-and-signal-store-features-3f094d9346d0).

I believe integrating this functionality into the core @ngrx/signals package (instead of relying on @angular-architects/ngrx-toolkit) would be highly beneficial for developers working with complex state management scenarios.

**Appreciation**
Thanks for all the great work you do @ngrx-toolkit! Looking forward to your thoughts on this.

Related [issue](https://github.com/ngrx/platform/issues/4707) already posted on @ngrx/platform

rainerhahnekamp:
> we often need to track the loading state of multiple collections simultaneously

The upcoming `resource` will define the standard for a status field in Angular. I think we should focus on integrating the `resource` into the SignalStore. Then these issues you've mentioned would be gone.

Would do you say?

rainerhahnekamp:
OK, but then please do the following before making a PR:
- Let's discuss the design here
- It should be backwards-compatible (no breaking changes)
- It needs to be supported by unit tests
- Documentation also has to be added.

abhijit-chikane:
Sure. I think the medium post I share already has pretty much everything explain.
It should be backwards-compatible (no breaking changes) => I think this is possible but needs some extra efforts.
I can help with the documentation.
@rainerhahnekamp 
Thanks for quick replies!😀

abhijit-chikane:
one question:

![Image](https://github.com/user-attachments/assets/6ed8803c-8214-4306-9bb9-612be75120d5)

Do we really need the config -> collection hierarchy in the arguments? Wouldn't it be simpler to just allow string | string[] instead?

[...]

PR #148, opened 2025-02-16
feat: Enhanced `withCallState` to accept mutliple collection configuration

abhijit-chikane:
feat: Enhanced `withCallState` to accept mutliple collection configuration
For #147 
I have not tested it but happy to help if in case it didn't work as expected
thanks

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-call-state.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-call-state.spec.ts
@@ -37,5 +37,6 @@ describe('withCallState', () => {
     expect(dataStore.entitiesCallState()).toBe('loaded');
     expect(dataStore.productsCallState()).toBe('loaded');
     expect(dataStore.entitiesLoaded()).toBe(true);
+    expect(dataStore.productsLoaded()).toBe(true);
   });
 });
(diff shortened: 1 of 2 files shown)


abhijit-chikane:
This requires additional changes and need to fix the typescript errors

abhijit-chikane:
@rainerhahnekamp did you get time to review this?
Let me know if anything needs to be done from my side
Thanks!

rainerhahnekamp (changes requested):
Well done and sorry for the late reply. What I'm currently asking myself, if we could use `collections` instead of `collection` to support the array of strings?

Would that be a big issue?

abhijit-chikane:
> Well done and sorry for the late reply. What I'm currently asking myself, if we could use `collections` instead of `collection` to support the array of strings?
> 
> Would that be a big issue?

Well initially I did exactly that but then it would be a breaking change
That’s why reverted back to collection

rainerhahnekamp:
>  it would be a breaking change

In which regard? To me, it would be an additional property. It would be breaking if `collections` was available in the past and now got its behavior changed.

rainerhahnekamp (approved):
LGTM

rainerhahnekamp:
Thanks @abhijit-chikane

PR #151, opened 2025-02-20
feat(ngrx-toolkit): Creates provideDevtoolsConfig

### LINKED ISSUE #149: Feature Request: Create injection token to replace "NgRx SignalStore" when calling connect
I would like to have the ability to configure the name of the application that appears in the Redux DevTools extension ("NgRx SignalStore"). Currently there isn't a way to change the name of the application after `window.__REDUX_DEVTOOLS_EXTENSION__.connect` is called. Since the `DevtoolsSyncer` already is an injectable service, please create an injection token that can be used to modify the app name when calling connect.

Example:
```
export const NGRX_DEVTOOLS_APPLICATION = new InjectionToken<string>('NgRxDevtoolsApplication');
```

mikerentmeister:
Creates a provideDevtoolsConfig function to allow the configuration of the Redux Devtools Extension.
fixes #149


docs/docs/provide-devtools-config.md
```diff
@@ -0,0 +1,64 @@
+---
```
rainerhahnekamp, line 1: this will be part of `with-redux.md`. You can add an additional chapter there, where you mention the possibility. Please use `bootstrapApplication` in the example instead of an `NgModule`.

rainerhahnekamp (changes requested):
Thanks for the PR. I've left a few notes.

mikerentmeister:
@rainerhahnekamp - Thank you for the feedback. I've pushed a new commit to address your comments.

libs/ngrx-toolkit/src/lib/devtools/tests/provide-devtools-config.spec.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  provideDevtoolsConfig,
+  REDUX_DEVTOOLS_CONFIG,
+  ReduxDevtoolsConfig,
+} from '../provide-devtools-config';
+
+describe('provideDevtoolsConfig', () => {
```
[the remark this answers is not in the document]
mikerentmeister: This is done

docs/docs/with-redux.md
```diff
@@ -107,3 +107,46 @@ signalStore(
...
+```typescript
+// app.config.ts
+import { ApplicationConfig } from '@angular/core';
+import { provideDevtoolsConfig } from '@angular-architects/ngrx-toolkit';
+
+export const appConfig: ApplicationConfig = {
+  providers: [
+    provideDevtoolsConfig({
+      name: 'MyApp',
+      maxAge: 50,
```
rainerhahnekamp, line 130: Needs to be removed here as well.

docs/docs/with-redux.md
```diff
@@ -107,3 +107,46 @@ signalStore(
...
+### Options
+
+The `provideDevtoolsConfig` function accepts an object with the following properties:
+
+- `name` (string): Optional name for the DevTools instance. The default is "NgRx SignalStore".
+- `maxAge` (number): Maximum number of actions to keep in the history. The default is 50.
+
+### Additional Information
+
+For more details on the available options and their usage, refer to the [Redux DevTools Extension documentation](https://github.com/zalmoxisus/redux-devtools-extension).
```
rainerhahnekamp, line 152: We should use the right link: https://github.com/reduxjs/redux-devtools

docs/docs/with-redux.md
```diff
@@ -107,3 +107,46 @@ signalStore(
...
+
+await bootstrapApplication(AppComponent, appConfig);
+```
+
+### Options
+
+The `provideDevtoolsConfig` function accepts an object with the following properties:
+
+- `name` (string): Optional name for the DevTools instance. The default is "NgRx SignalStore".
+- `maxAge` (number): Maximum number of actions to keep in the history. The default is 50.
```
rainerhahnekamp, line 148: No `maxAge`. We don't support it. That would be a separate issue.

docs/docs/with-redux.md
```diff
@@ -107,3 +107,46 @@ signalStore(
...
+
+// main.ts
+import { bootstrapApplication } from '@angular/platform-browser';
+import { AppComponent } from './app/app.component';
+import { appConfig } from './app/app.config.ts';
+
+await bootstrapApplication(AppComponent, appConfig);
+```
+
+### Options
```
rainerhahnekamp, line 143: Since we have only `name`, I guess we can remove the options.

docs/docs/with-redux.md
```diff
@@ -107,3 +107,46 @@ signalStore(
   })
 );
 ```
+
+### Configuring the Redux Devtools Extension
+
+The `provideDevtoolsConfig` function allows you to configure the Redux DevTools integration for your NgRx SignalStore. This function is essential for setting up the DevTools with custom options. The function only needs to be called once in your appConfig or AppModule.
+
+### Usage
```
rainerhahnekamp, line 115: No need for a separate header?

libs/ngrx-toolkit/src/lib/devtools/provide-devtools-config.ts
```diff
@@ -0,0 +1,62 @@
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+/**
+ * Provides the configuration options for connecting to the Redux DevTools Extension.
+ */
+export function provideDevtoolsConfig(
```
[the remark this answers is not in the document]
rainerhahnekamp: At the moment, we can't really support anything of the options of the original. It would be great if we get them, but not in this PR.

libs/ngrx-toolkit/src/lib/devtools/provide-devtools-config.ts
```diff
@@ -0,0 +1,42 @@
...
+);
+
+/**
+ * Options for connecting to the Redux DevTools Extension.
+ *
+ * This interface represents the configuration options that can be passed
+ * to the Redux DevTools Extension's connect function. It includes options
+ * for enabling features, serialization, tracing, action creators, and other
+ * settings that control the behavior and appearance of the DevTools.
+ * See {@link https://github.com/reduxjs/redux-devtools/blob/main/extension/docs/API/Arguments.md the Redux DevTools API documentation}.
```
rainerhahnekamp, line 29: We don't support them yet.

rainerhahnekamp (changes requested):
I see that you would like to support the options of the Devtools.

What about the following suggestion. We do this PR just with name, and then we can check in a separate one, would options are supported automatically by the DevTools and for which options we need to provide additional implementation?

mikerentmeister:
@rainerhahnekamp - Thanks for the great suggestions again. I have updated the PR to align with your comments.

rainerhahnekamp:
1000 thanks @mikerentmeister. Expect a new release of the toolkit with your feature soon!

PR #160, opened 2025-03-14
reset only subset of the state in `withReset()`

daboudii:
In case you need to re initialize your state

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-reset.ts
+++ b/libs/ngrx-toolkit/src/lib/with-reset.ts
@@ -10,6 +10,7 @@ import {
 
 export type PublicMethods = {
   resetState(): void;
+  resetSlice(slice: string | string[]): void;
 };
 
 /**
@@ -20,12 +21,23 @@ export type PublicMethods = {
  */
 export function withReset() {
   return signalStoreFeature(
-    withProps(() => ({ _resetState: { value: {} } })),
+    withProps(() => ({ _resetState: { value: {} as Record<string, any> } })),
     withMethods((store): PublicMethods => {
       // workaround to TS excessive property check
       const methods = {
         resetState() {
           patchState(store, store._resetState.value);
+        },
+        resetSlice(slice: string | string[]) {
+          patchState(store, (state) => ({
+            ...state,
+            ...(typeof slice === 'string'
+              ? { [slice]: store._resetState.value[slice] }
+              : slice.reduce((acc, key) => {
+                  acc[key] = store._resetState.value[key];
+                  return acc;
+                }, {} as Record<string, any>)),
+          }));
         },
         __setResetState__(state: object) {
           store._resetState.value = state;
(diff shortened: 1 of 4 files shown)


rainerhahnekamp:
Thanks @daboudii, there is already an extension `withReset`. So I would suggest to use that one and if there is something missing, to make a PR for `withReset`.

I have to admit that I like your approach of using a function `setInitial()` in `patchState` much, much better than how it is implemented in `withReset`. So maybe that is something you could start with - if you want.

daboudii:
The `withReset` do not meet my needs since it reset all the store and not just the callState, I implemented the `setInitial()` in this MR, is there anything more to add ?

rainerhahnekamp:
@daboudii, yes what options do we have to do this in `withReset()`? 

Would it help, if `withReset()` supports selecting a subset of the state?

The reason why I'm pushing for the integration in the other feature is that we would not be limited to support a very specific use case ( `callState`) but a much wider range.

daboudii:
@rainerhahnekamp  let me know if it need something else

libs/ngrx-toolkit/src/lib/with-reset.ts
```diff
@@ -20,13 +21,24 @@ export type PublicMethods = {
...
-    withProps(() => ({ _resetState: { value: {} } })),
+    withProps(() => ({ _resetState: { value: {} as Record<string, any> } })),
     withMethods((store): PublicMethods => {
       // workaround to TS excessive property check
       const methods = {
         resetState() {
           patchState(store, store._resetState.value);
         },
+        resetSlice(slice: string | string[]) {
+          patchState(store, (state) => ({
```
abhijit-chikane, line 32: It would be even better if we could just reset the part of the slice
in case of obj sometimes we don't want to reset the whole inner object but part of the object

daboudii: Something like `resetSlice('address.city')` ?

abhijit-chikane: > Something like `resetSlice('address.city')` ? If yes it will add even more complexity, is it really necessary ?

I think it will definitely add complexity but it will add much more value for project like ours where data structures are much more complex

daboudii: I don't think I will add that, in this case just easier to patch your state with the value you which to change.

rainerhahnekamp (changes requested):
Good job! I have a question about the design.

State changes would be better handled as functions within patchState. This ensures that patchState remains the sole entity modifying state and allows multiple update functions to run together as an “atomic update” of the internal Signal.

That being said, `setResetState()` was already implemented in a way that doesn’t align with this approach. What if we provide a more flexible `reset()` function that supports both full and partial state resets and is type-safe as well.

```typescript
// reset the full state
patchState(store, reset())

// reset the partial state
patchState(store, reset(state => ({address: state.address, user: state.user}))
```

What do you say?

libs/ngrx-toolkit/src/lib/with-reset.ts
rainerhahnekamp: I touched upon that in my review below https://github.com/angular-architects/ngrx-toolkit/pull/160#pullrequestreview-2694431498

daboudii:
> What do you say?

Yes it looks way better like that !

rainerhahnekamp:
Closing that one in favour of #204

Issue #155, opened 2025-03-11
docs: add import to doc page


michael-small:
It is not that uncommon for intelisense in whatever IDE or Stackblitz to be lacking. For example, I just wanted to use `withFeatureFactory` and I had to pull up the existing import I used in a different project to know it is `import { withFeatureFactory } from '@angular-architects/ngrx-toolkit'`. 

I suggest that the docs include an import (and ideally with a click to copy), like how ngxtension and other Angular libraries include imports at the top of a util's page

> https://ngxtension.netlify.app/utilities/directives/merge-inputs/
> import { mergeInputs } from 'ngxtension/inject-inputs';

I would be willing to handle the PR if I have the greenlight for this. Not sure if click to copy is possible out of the box with the docs but if it can be handled with JS then I would do that as well.

michael-small:
Accomplished this a bit ago and forgot to close

PR #162, opened 2025-03-15
docs: add `import` statements to doc pages

michael-small:
Issue: docs: add import to doc pages #155

https://github.com/angular-architects/ngrx-toolkit/issues/155

### DIFF
--- a/docs/docs/create-redux-state.md
+++ b/docs/docs/create-redux-state.md
@@ -2,6 +2,10 @@
 title: createReduxState()
 ---
 
+```typescript
+import { createReduxState } from '@angular-architects/ngrx-toolkit/redux-connector';
+```
+
 The Redux Connector it is not an extension but turns any `signalStore()` into a Global State Management Slice following the Redux pattern.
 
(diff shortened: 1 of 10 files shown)


michael-small:
@rainerhahnekamp I added imports to the big functions on the sidebar. However, there is helper functions for some things, such as

```typescript
import { renameDevtoolsName } from '@angular-architects/ngrx-toolkit';
```

or

```typescript
import { createReducer } from '@angular-architects/ngrx-toolkit';
```

among some other functions.

I assume each should have its import listed somewhere on the pages too? I'm thinking either

1. In the example blocks, or
2. Under its own header just like the whole pages have their own block at the top

Thoughts? 

Once you make a call on that I'll be ready to get this out as a ready PR

rainerhahnekamp:
Can we add it to both or would that be too much?

msmallest:
That should be good

michael-small:
I went back across everything all the pages and added imports to everything including

- Top of the page for the main functions
- Top of the subsections for helper functions
- All code blocks using any toolkit function, even if the function isn't the main example (`withUndoRedo` in `withDataService` examples etc)

I also accidentally managed to add imports for `withReset` even though that is the other PR, but that PR still adds it to the sidenav so that PR is still relevant.

rainerhahnekamp (approved):
Thanks a lot @michael-small

PR #163, opened 2025-03-15
docs: add `withReset` to sidenav + add imports to `withReset`

michael-small:
I saw `withReset` was not on the sidenav

Adjacent to the imports at the top of the page PR

- I added an import like in the other PR but in this branch
- Setting this as a draft as my open question in the imports PR applies to `setResetState()`

### DIFF
--- a/docs/docs/with-reset.md
+++ b/docs/docs/with-reset.md
@@ -2,6 +2,10 @@
 title: withReset()
 ---
 
+```typescript
+import { withReset } from '@angular-architects/ngrx-toolkit';
+```
+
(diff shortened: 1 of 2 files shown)


michael-small:
The other docs PR where I added imports accidentally added `withReset` anyways, but this PR still will include addition to the sidenav so it still has a purpose.

michael-small:
I have marked this PR as ready.

With respect to @daboudii's point, I see there is not a dedicated docs page for `withCallState`. So in my opinion in relation to this PR, that's its own topic for a separate issue/PR of its own but thanks for pointing that out.

rainerhahnekamp (approved):
LGTM. Thanks again.

Issue #175, opened 2025-05-27
withCallState does not work as described in the docs


justalemming:
As described here: [https://ngrx-toolkit.angulararchitects.io/docs/with-call-state#helper-methods](https://ngrx-toolkit.angulararchitects.io/docs/with-call-state#helper-methods) a signalStore with the withCallState() feature should have the methods store.setLoading(), store.setLoaded() and store.setError(). These are not available.

eminvegan:
**Yes, same here. Getting error and even Types are not available in the store:**

<img src="https://github.com/user-attachments/assets/0861ae1b-01d4-4ccc-825b-44783fe8e4b0" width="480"><br /><br />

### Code

```typescript
export const FileListStore = signalStore(
  { providedIn: 'root' },
  withDevtools('FileListComponent'),
  withState(initialState),
  withCallState(),

  withProps(() => ({
    authenticationStore: inject(AuthenticationStore),
    firestore: inject(Firestore),
    driveApiService: inject(DriveApiService),
    fileListService: inject(FileListService),
    groupListItemStore: inject(GroupListItemStore),
    googleApiProviderService: inject(GoogleApiProviderService),
  })),

  withEntities(fileListPageTokenEntityConfig),
  withEntities(fileListEntityConfig),

  withComputed(
    ({
      selectedClientGroupId,
      selected,
      pageSize,
      pageIndex,
      orderByField,
      sortDirection,
      pageLastElements,
    }) => ({
      loadPageDataParams: computed(() => {
        return {
          pageDataSelectedClientGroupId: selectedClientGroupId(),
          pageDataSelected: selected(),
          pageDataPageSize: pageSize(),
          pageDataPageIndex: pageIndex(),
          pageDataOrderByField: orderByField(),
          pageDataSortDirection: sortDirection(),
          pageDataPageLastElements: pageLastElements(),
        };
      }),
    }),
  ),

  withMethods(
    ({
      authenticationStore,
      firestore,
      fileListService,
      googleApiProviderService,
      ...store
    }) => ({
      updateSelectedClientGroup(value: ClientGroup): void {
        updateState(store, '[FileListComponent] Update Selected Client Group', {
          selectedClientGroup: value,
          selectedClientGroupId: value.id,
        });
      },
      updateTotal(value: number): void {
        updateState(store, '[FileListComponent] Update Total', {
          total: value,
        });
      },
      updatePageSize(value: number): void {
[...]

PR #176, opened 2025-05-28
feat(with-call-state): add helper methods for updaters

gremlin896:
Fixes https://github.com/angular-architects/ngrx-toolkit/issues/175 by adding the documented feature that was missing from the implementation.

Alternatively the documentation can be updated instead to correctly describe the functionality.


rainerhahnekamp:
I would rather not add those helper methods, but stick only to the standalone functions, as it is now.

The more methods generic features add, the higher the risk of overloading methods and props for the consumers.

So, in the end, it would be better to fix the docs.

gremlin896:
Agreed, I'll close this off

PR #177, opened 2025-05-28
docs(with-call-state): fix custom updaters documentation

### LINKED ISSUE #175: withCallState does not work as described in the docs

gremlin896:
Fixes #175 by updating the docs to match the implementation.

As noted on #176 the preferred approach is to rely on the built-in `patchState`, rather than adding methods that wrap them.

### DIFF
--- a/docs/docs/with-call-state.md
+++ b/docs/docs/with-call-state.md
@@ -14,33 +14,16 @@ The simplest way to use `withCallState` is without any configuration:
 
 ```typescript
 export const TodosStore = signalStore(
-  withCallState(),
+  withCallState()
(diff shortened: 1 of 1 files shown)


rainerhahnekamp (approved):
Wonderful, thanks Matthew!

Issue #181, opened 2025-06-26
v20 Roadmap


rainerhahnekamp:
# v20 Roadmap

NgRx-Toolkit will support NgRx & Angular **20** as soon as NgRx ships the pre-release for v20.

## Next steps

There will be a final release for version **19**, which includes the following changes:

- `withLocalStorageSync` will support `indexedDb` in a non-breaking change  
- All commits and changes added after `v19.1.0` will be included  
- Support for `ng update`  

---

A pre-20 release of the NgRx Toolkit will be published with:

- A dependency on the pre-release of NgRx 20

---

## Planned features for Toolkit v20

These features are planned for Toolkit v20 and may be released across multiple versions.  
**These can also be implemented via community contributions.**

- Add `withResource` as an incubator feature, based on the current [design and PR](https://github.com/ngrx/platform/issues/4833)  
- Once `withResource` is released in `@ngrx/signals`, the toolkit will provide an automatic migration script  
- Explore and implement support for `@ngrx/signals/events` in `withDevtools`  
- Extend `withStorageSync` to support browser tab synchronization  
- Add an API for asynchronous/synchronous storage synchronization  
  - _for example_, to sync with a backend service

PR #183, opened 2025-07-02
feat: added selectId

### LINKED ISSUE #67: How to use DataService when my entity has no 'id' property?
Is it possible to use an adapter to map my "id" property to "someId"?

DrakeAnglin:
close #67

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-data-service.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-data-service.spec.ts
@@ -1,9 +1,8 @@
 import { Injectable } from '@angular/core';
 import { fakeAsync, TestBed, tick } from '@angular/core/testing';
-import { Observable, firstValueFrom, of, delay } from 'rxjs';
 import { signalStore, type } from '@ngrx/signals';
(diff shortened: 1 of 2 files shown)


rainerhahnekamp:
If it has no bugs, it will make it also into the 19 branch. For more details: #181

libs/ngrx-toolkit/src/lib/with-data-service.spec.ts
```diff
@@ -414,6 +576,19 @@ const createFlight = (flight: Partial<Flight> = {}) => ({
   },
   ...flight,
 });
+
+let currentCustomFlightId = 0;
+const createFlightWithCustomId = (
+  flight: Partial<FlightWithCustomId> = {}
+): FlightWithCustomId => ({
+  flightId: `${++currentCustomFlightId}`,
```
rainerhahnekamp, line 584: let's to the conversion how it is also suggested in the Google styleguide.
```suggestion
  flightId: String(++currentCustomFlightId),
```

rainerhahnekamp (changes requested):
Looks good so far. Can you also extend the current documentation so that it includes the `selectId` option?

Issue #185, opened 2025-07-09
[undo/redo]: clearing the stack keeps previous item in the undo-stack


GregOnNet:
Hey together,

I observed something unexpected, after clearing the undo-redo-stack.
The `previous` item is not set to `null` and will be added to the undo-stack again.

The fix seems to be easy.
It works in my project, at least. :P

```ts
 clearStack(): void {
        undoStack.splice(0);
        redoStack.splice(0);
        previous = null;     // 👈 clearing previous

        updateInternal();
      }
```

I would like to contribute a PR, @rainerhahnekamp .

Cheers
Gregor

PR #186, opened 2025-07-09
fix(undo-redo): clearStack should also clear the previous item

GregOnNet:
solves #185

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
@@ -210,5 +210,36 @@ describe('withUndoRedo', () => {
       expect(store.canUndo()).toBe(false);
       expect(store.canRedo()).toBe(false);
     })
+
+    it('cannot undo after clearing and setting a new value', fakeAsync(() => {
+      const Store = signalStore(
+        { providedIn: 'root' },
+        withState(testState),
+        withMethods((store) => ({
+          update: (value: string) => patchState(store, { test: value }),
+        })),
+        withUndoRedo({ keys: testKeys })
+      );
+
+      const store = TestBed.inject(Store);
+
+      store.update('Alan');
+      tick(1);
+
+      store.update('Gordon');
+      tick(1);
+
+      store.clearStack();
+      tick(1);
+
+      // After clearing the undo/redo stack, there is no previous item anymore.
+      // The following update becomes the first value.
+      // Since there is no other value before, it cannot be undone.
+      store.update('Hugh');
+      tick(1);
+
+      expect(store.canUndo()).toBe(false);
+      expect(store.canRedo()).toBe(false);
+    }));
   });
 });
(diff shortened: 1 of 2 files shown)


libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
```diff
@@ -22,39 +32,55 @@ describe('withUndoRedo', () => {
         'canRedo',
         'undo',
         'redo',
-        'clearStack'
+        'clearStack',
```
GregOnNet, line 35: `withUndoRedo` is used as plugin inside a store.
When the undo & redo stack shall be cleared `clearStack` is called.

Currently, it is not clear which stack is cleared.
In fact, we are clearing two stacks.

Do we want to deprecate `clearStack` in favor of `resetUndoRedo` or `clearUndoRedo`?
I guess `reset*` fits better, @rainerhahnekamp .

rainerhahnekamp: Hey @GregOnNet, regarding the reset functions: there’s an open discussion about taking a different approach. The idea is to avoid adding more methods directly to the signalStore, in order to prevent “polluting” the API with keywords that might be needed for other use cases in the future:
https://github.com/angular-architects/ngrx-toolkit/pull/160#pullrequestreview-2694431498

So yes, I’m in favour of deprecating clearStack — but instead of introducing new store methods, we should provide standalone functions that can be used within patchState.

GregOnNet: When I understand correctly clearing the undo-redo-stack could become something like this.

```ts

patchState(store, state => reset({ undoStack: [], redoStack: []}));

// or if the internal state gets adjusted

patchState(store, state => reset({ undoRedo: ??? }));

// or with some type-literal wizardry

patchState(store, state => reset('undoRedo'));
```

**Am I correct?**

I also experience challenges managing the visibility of certain mechanics in the store.
It feels like there is "internal state" and "public state".
In consequence there are methods internal to the store and public ones.

Nevertheless, I do not think that the example `clearStack` is "polluting" the API.
To me, it is the public API of the `signalStoreFeature`.
So I absolutely expect this method to be in my store to handle my use-cases.
Adding `reset`, causes an indirection, a developer might need to read about.
To me, it is not intuitive.
If I understood correctly, instead of "learning" about "clearStack" the API-user now needs to have knowledge about the shape of the state of the undo-redo-plugin.

Maybe it is worth to push the discussion about visibility to @ngrx/signal-store instead to reason about an overall solution for this topic.

GregOnNet:
I pushed the changes withoud reformatting @rainerhahnekamp .

rainerhahnekamp:
@GregOnNet, thanks, and a quick update from my side. I am currently on vacation and will return on Wednesday. You will hear from me then.

rainerhahnekamp (approved):
Thanks @GregOnNet

GregOnNet:
Hi @rainerhahnekamp ,

- Prettier: sure, why not. :-)
- Standalone `reset`-function. It's a pleasure to help with this, too. We would need an issue to clarify the shape of the API, since there are 2 directions, we discussed (see https://github.com/angular-architects/ngrx-toolkit/pull/186#discussion_r2196548590).
  - Can you please outline the goal and the API-spec in a short issue, please?

PR #192, opened 2025-07-26
feat: add with-connect feature

manfredsteyer:
withConnect() adds a method to connect Signals back to the store. Everytime these Signals change, the store will be updated accordingly.


michael-small (commented):
One nitpick: I would find + replace 

`poeple`

with

`people`

Good stuff, this is a great idea.

rainerhahnekamp:
I am closing that in favor of `delegatedSignal`/`linkedSignal`.

In hindsight, we should have implemented `withConnect`. Users would have had the option to use it for almost one year now.

PR #202, opened 2025-08-08
feat: add mutation api

manfredsteyer:
rxMutation creates a mutation function in the store, delegating to a function/ method returning an observable, and takes care of race conditions by internally using mergeMap (default), switchMap, concatMap, or exhaustMap. 

mutation also creates an async mutation function, delegating to a function/ method returning an observable. When the async function is awaited without throwing an exception, the caller can safely assume that it has been executed. However, here, the consumer is responsible for preventing race conditions.


libs/ngrx-toolkit/src/lib/with-pagination.spec.ts
```diff
@@ -42,7 +42,6 @@ describe('withPagination', () => {
     );
 
     const store = new Store();
-
     patchState(
```
rainerhahnekamp, line 45: that file could be reverted.

libs/ngrx-toolkit/src/lib/rx-mutation.ts
```diff
@@ -0,0 +1,171 @@
...
+          param,
+          resolve,
+        });
+      }
+    });
+  };
+
+  const mutation = mutationFn as Mutation<P, R>;
+  mutation.status = status;
+  mutation.callCount = callCount;
```
rainerhahnekamp, line 167: Can we remove that one now? If we do it later for standalone, it would be breakign change.

libs/ngrx-toolkit/src/lib/with-mutations.spec.ts
```diff
@@ -0,0 +1,537 @@
...
+    const store = testSetup.store;
+
+    expect(store.incrementStatus()).toEqual('idle');
+    expect(store.incrementProcessing()).toEqual(false);
+
+    store.increment(2);
+    expect(store.incrementStatus()).toEqual('processing');
+    expect(store.incrementProcessing()).toEqual(true);
+
+    tick(2000);
```
rainerhahnekamp, line 130: A todo for the refactoring. We don't want tests that artificially extend the duration if it is not absolutely necesary.

libs/ngrx-toolkit/src/lib/with-mutations.spec.ts
```diff
@@ -0,0 +1,537 @@
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
```
rainerhahnekamp, line 1: TODO: Generally speaking, once we make `rxMutation` standalone, most tests of `rxMutation` will cover already what we have here.

libs/ngrx-toolkit/src/lib/with-mutations.ts
```diff
@@ -0,0 +1,169 @@
...
+export function withMutations<
+  Input extends SignalStoreFeatureResult,
+  Result extends MutationsDictionary,
+>(
+  mutationsFactory: (
+    store: Input['props'] & Input['methods'] & StateSource<Input['state']>,
+  ) => Result,
+): SignalStoreFeature<Input> {
+  return (store) => {
+    // TODO: Is this the correct usage?
```
rainerhahnekamp, line 118: We could leave it is. Might be better options, but we since we are not importing any internal stuff here, it is good to go.

libs/ngrx-toolkit/src/lib/with-mutations.ts
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+export type MutationStatus = 'idle' | 'processing' | 'error' | 'success';
+
+// withMethods uses Record<string, Function> internally
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+export type MethodsDictionary = Record<string, Function>;
+
+type NamedMutationProps<T extends MutationsDictionary> = {
+  [Prop in keyof T as `${Prop & string}Processing`]: Signal<boolean>;
```
rainerhahnekamp, line 45: I would rename it to `IsProcessing`. `incrementerIsProcessing` might be easier to read than `incrementerProcessing`.

libs/ngrx-toolkit/src/lib/flattening-operator.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type RxJsFlatteningOperator = <T, O extends ObservableInput<unknown>>(
+  project: (value: T, index: number) => O,
+) => OperatorFunction<T, ObservedValueOf<O>>;
+
+/**
+ * A wrapper for an RxJS flattening operator.
+ * This wrapper informs about whether the operator has exhaust semantics or not.
+ */
+export type FlatteningOperator = {
```
rainerhahnekamp, line 19: TODO: Works for now but we should think about something more generic than `exhaustSemantic` boolean for custom flatteners. The only thing we need to know, is if an operator skips an input completely. That means it doesn't even subscribe to it.

rainerhahnekamp (changes requested):
This review consists of two parts. Things that should be addressed before we merge it, and things that can be done later, as additional features.

I've added a `TODO` prefix to all my comments that fall into the second category (which could be done later).

Here's a summary, where we could have separate Issues and PR contributions.

- #220 `rxMutate` as fully standalone function which would also work outside the SignalStore
- Documentation in Docusaurus
- Refactoring of Tests (depends on standalone `rxMutation`, we can do more directly in that file and don’t have to go through the complete SignalStore.
- Allow `withMutations` to expose more tracking data of the mutation, like `isPending`, `hasValue`.
- E2E Tests for Mutation Demo
- Standalone and Promise-based `mutation`
- #221 Standalone `httpMutation` in the same manner as `httpResource`.
- Redesign of flatten operator: Should probably have a callback telling that a certain input has been skipped.

libs/ngrx-toolkit/src/lib/rx-mutation.ts
manfredsteyer: Done

libs/ngrx-toolkit/src/lib/with-mutations.ts
```diff
@@ -0,0 +1,169 @@
...
+      status: 'aborted';
+    };
+
+export type MutationStatus = 'idle' | 'processing' | 'error' | 'success';
+
+// withMethods uses Record<string, Function> internally
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+export type MethodsDictionary = Record<string, Function>;
+
+type NamedMutationProps<T extends MutationsDictionary> = {
```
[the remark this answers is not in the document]
rainerhahnekamp: Will be done in a separate Issue

libs/ngrx-toolkit/src/lib/flattening-operator.ts
rainerhahnekamp: Will be done in a separate issue

libs/ngrx-toolkit/src/lib/with-mutations.spec.ts
```diff
@@ -0,0 +1,537 @@
...
+      store,
+      onSuccessCalls: () => onSuccessCalls,
+      onErrorCalls: () => onErrorCalls,
+      lastOnSuccessParams: () => lastOnSuccessParams,
+      lastOnErrorParams: () => lastOnErrorParams,
+    };
+  });
+}
+
+describe('withMutations with rxMutation', () => {
```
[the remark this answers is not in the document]
rainerhahnekamp: Separate Issue

PR #207, opened 2025-08-09
fix #190: make withUndoRedo sync with watchState instead of effect

### LINKED ISSUE #190: [undo/redo]: clearStack is executed before the store is modified
Hi,

if I execute clearStack directly in a store method after reverting / initializing a part of the store, the clearStack is executed before the store is modified and the undo/redo stack gets the change as initial value.

```
export const EditorStore = signalStore(
    { providedIn: 'root' },
    ...
    withUndoRedo({
        maxStackSize: 100,
        keys: ['selectedFile']
    }),
    withMethods((store) => {
      ...
      selectFile(file: ProjectFile) {
          patchState(store, {
              selectedFile: selected,
          });
    
          store.clearStack();
      },
      ...
});
```

The problem may be that clearStack is executed synchronously, whereas the undo/redo uses an effect that is executed asynchronously.

https://github.com/angular-architects/ngrx-toolkit/blob/399903d198cd8be3ff3ad0df4de5843de855167f/libs/ngrx-toolkit/src/lib/with-undo-redo.ts#L139C9-L139C15

The solution might be to use ```watchState``` instead of ```effect```, which is executed synchronously.

Am I right? Should I create a pull request with this fix?

richardliebmann:
This pull request fixes bug [#190](https://github.com/angular-architects/ngrx-toolkit/issues/190) and causes the undo/redo feature to synchronise instead of operating asynchronously.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-undo-redo.ts
+++ b/libs/ngrx-toolkit/src/lib/with-undo-redo.ts
@@ -1,10 +1,11 @@
-import { Signal, effect, isSignal, signal, untracked } from '@angular/core';
+import { Signal, isSignal, signal, untracked } from '@angular/core';
 import {
   EmptyFeatureResult,
   SignalStoreFeature,
   SignalStoreFeatureResult,
   patchState,
   signalStoreFeature,
+  watchState,
   withComputed,
   withHooks,
   withMethods,
@@ -143,7 +144,7 @@ export function withUndoRedo<Input extends EmptyFeatureResult>(
     })),
     withHooks({
       onInit(store) {
-        effect(() => {
+        watchState(store, () => {
           const cand = keys.reduce((acc, key) => {
             const s = (store as Record<string | keyof Input['state'], unknown>)[
               key
--- a/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-undo-redo.spec.ts
@@ -1,5 +1,5 @@
 import { computed, inject } from '@angular/core';
-import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { TestBed } from '@angular/core/testing';
 import {
   patchState,
   signalStore,
@@ -73,7 +73,7 @@ describe('withUndoRedo', () => {
   });
 
   describe('undo and redo', () => {
-    it('restores previous state for regular store key', fakeAsync(() => {
+    it('restores previous state for regular store key', () => {
       TestBed.runInInjectionContext(() => {
         const Store = signalStore(
           withState(testState),
@@ -85,24 +85,22 @@ describe('withUndoRedo', () => {
         );
 
         const store = new Store();
-        tick(1);
 
         store.updateTest(newValue);
-        tick(1);
+
         expect(store.test()).toEqual(newValue);
         expect(store.canUndo()).toBe(true);
         expect(store.canRedo()).toBe(false);
 
         store.undo();
-        tick(1);
 
         expect(store.test()).toEqual('');
         expect(store.canUndo()).toBe(false);
         expect(store.canRedo()).toBe(true);
       });
-    }));
+    });
 
-    it('restores previous state for regular store key and respects skip', fakeAsync(() => {
+    it('restores previous state for regular store key and respects skip', () => {
       TestBed.runInInjectionContext(() => {
         const Store = signalStore(
           withState(testState),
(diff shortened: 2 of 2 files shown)


jdegand:
@rainerhahnekamp You add something like this: 

```ts
withMethods((store) => ({
  undo(): void {
    const item = undoStack.pop();

    if (item && previous) {
      redoStack.push(previous);
    }

    if (item) {
      skipOnce = true;
      patchState(store, item);
      previous = item;
    }

    updateInternal();
  },
  redo(): void {
    const item = redoStack.pop();

    if (item && previous) {
      undoStack.push(previous);
    }

    if (item) {
      skipOnce = true;
      patchState(store, item);
      previous = item;
    }

    updateInternal();
  },
  clearStack(): void {
    undoStack.splice(0);
    redoStack.splice(0);
    previous = null;
    updateInternal();
  },
  softClearStack(): void {
    undoStack.splice(0);
    redoStack.splice(0);

    // Capture current state as new baseline
    const cand = keys.reduce((acc, key) => {
      const s = (store as Record<string | keyof Input['state'], unknown>)[key];
      if (s && isSignal(s)) {
        acc[key] = s();
      }
      return acc;
    }, {} as StackItem);

    previous = cand;
    updateInternal();
  },
}))
```

to match the `soft clear` functionality you are testing for.

GregOnNet:
Hi @rainerhahnekamp,

the test you brought up is worth discussing it.
However, it is not related to this PR but to https://github.com/angular-architects/ngrx-toolkit/pull/186.

The behaviour of `clearStack` is debatable.
Maybe another issue is a better place for it.

For my use case, `clearStack` is like starting all over again (initial state).
Therefore, we need to forget about the complete history.

Here is a table that describes the flow of your test, Rainer.

|state|previousValue|operation|canUndo|comment|
|-|-|-|-|-|
|0|null|**initial value**| false||
|1|0|updateNumber| true||
|0|null|undo|false||
|10|0|updateNumber|true||
|10|null|clearStack|false|_this is like enforcing the_ **initial value**|
|11|null|updateNumber|false|_behaves like_ **initial value**|

## Use Case

Imagine you have a form-editor that can load and edit different forms.
You can navigate between those forms.
You want to undo/redo changes to several forms.
Once you navigate to another form, you need to call `clearStack` and start all over with the next form.
We need to ensure that nothing from the previous form is present in the undo/redo-stack.
Otherwise, we would undo a change in the new form and would get the previous form-state.

libs/ngrx-toolkit/src/lib/with-undo-redo.ts
```diff
@@ -143,7 +144,7 @@ export function withUndoRedo<Input extends EmptyFeatureResult>(
     })),
     withHooks({
       onInit(store) {
-        effect(() => {
+        watchState(store, () => {
```
GregOnNet, line 147: This change is great and allows us to make all tests in `with-und-redo.spec.ts` synchronous.

Can we omit all `fakeAsync`- and `tick`-usages there?

GregOnNet (changes requested):

richardliebmann:
Hi @GregOnNet and @rainerhahnekamp,

I have removed the asynchronous features from the test, including 'fakeAsync' and 'tick'.

In my opinion, @GregOnNet is right. We also use the clearStack feature to reset to the initial value. In our case, we call clearStack after selecting or loading a new file in an editor and don't want to go back to the previous file. I currently cannot imagine a requirement where softClearStack would make sense, but it might be a useful extension.

rainerhahnekamp:
@GregOnNet, @richardliebmann, sorry but I still don't get it.

The core issue is that during initialization the stack always gets the initial state but not on `clearStack`().

As a result, the first change after clearing (e.g. setting `value` to `11`) is treated as the **new initial state**, and there’s nothing to undo until the second change.

Using your editors analogy: If you move from one file to another, the second file also has an initial state, and that’s exactly what the stack should adopt as its baseline.

Pseudo code for Richard’s case:
```ts
store.selectNewFile('graz.jpg');

// store's state is cleared and becomes the initial state for Graz
store.clearStack(); // Graz file's state is now the baseline

store.setTitle('Wien');

store.undo(); // restores title back to Graz
```
---

@jdegand This is not about the actual watchState or other implementation details - it’s purely about defining the intended behavior of the feature.

GregOnNet:
Hi @rainerhahnekamp,

@richardliebmann please correct me, if I am wrong.
From my understanding, the order of execution was not as expected, since undo/redo “was” async because of `effect`.

If you write…

```ts
patchState(...);
store.clearStack();
```

… it is possible that `clearStack`, was executed before the change was added to the undo-Stack. Although we expect `canUndo` to be `false` in this case, it was true.

rainerhahnekamp:
@GregOnNet Yes, this is addressing an issue unrelated to Richard’s original intention. I came across it while writing a test for this particular change.

Should we merge this as is and open a separate issue for the new problem?

GregOnNet:
Yes, this PR can be merged.

Concerning the new problem, we might want to jump on a short call.
I am a bit confused what the new issue is.
Maybe the description in the new issue makes things clearer to me. :-)

PR #211, opened 2025-08-12
chore(demo app): cleanup appearance and other code

michael-small:
The demo application is my go-to way to test things live. When running it recently, I found and fixed the following:

- Fixed sidenav being hidden by incompatible CSS
- Alphabetized sidenav by feature
- General cleanup of `withReset` page, as I was experimenting with changes in that feature
- I will probably revamp more features like the `withReset` cleanup before marking this for review

Screenshot from 2025-08-12 01-50-57

### DIFF
--- a/apps/demo/src/app/app.component.html
+++ b/apps/demo/src/app/app.component.html
@@ -1,7 +1,7 @@
 <mat-toolbar color="primary">
   <span>NgRx Toolkit Demo</span>
 </mat-toolbar>
-<mat-drawer-container class="container">
+<mat-drawer-container>
   <mat-drawer mode="side" opened>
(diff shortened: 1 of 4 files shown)


jdegand:
Below `768px`, the app is unusable.  Have you looked at responsive changes?

michael-small:
Good catch, I have been focused on it being a bit broken in a more traditional desktop size/ratio but had not considered smaller sizes. Will check that out once I get some stuff at the size I was working with on par with before.

michael-small:
> Below `768px`, the app is unusable. Have you looked at responsive changes?

@jdegand I made a separate change to make the sidenav collapsible

PR #214, opened 2025-08-21
`withResource`

rainerhahnekamp:


docs/docs/with-resource.md
```diff
@@ -0,0 +1,121 @@
+---
+title: withResource()
+---
+
+```typescript
+import { withResource } from '@angular-architects/ngrx-toolkit';
+```
+
+> **⚠️ Important Note**: This extension is very likely to land in NgRx once Angular's `Resource` gets developer preview. The `withResource` extension provides early access to this functionality and will be maintained for compatibility until the official NgRx implementation is available.
```
michael-small, line 9:
```suggestion
> **⚠️ Important Note**: This extension is very likely to land in NgRx once Angular's `Resource` enters developer preview. The `withResource` extension provides early access to this functionality and will be maintained for compatibility until the official NgRx implementation is available.
```

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
```diff
@@ -0,0 +1,662 @@
...
+
+      describe('override protection', () => {
+        const warningSpy = jest.spyOn(console, 'warn');
+
+        afterEach(() => {
+          warningSpy.mockClear();
+        });
+
+        for (const memberName of [
+          //TODO wait for https://github.com/ngrx/platform/pull/4932
```
michael-small, line 380: Will this PR waiting for this to land in the store first? Or is this just a note to keep in mind that it will eventually be ready.

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
```diff
@@ -0,0 +1,662 @@
...
+
+            expect(warningSpy).toHaveBeenCalledWith(
+              '@ngrx/signals: SignalStore members cannot be overridden.',
+              'Trying to override:',
+              memberName,
+            );
+          });
+        }
+
+        //TODO wait for https://github.com/ngrx/platform/pull/4932
```
michael-small, line 407: Same question as above - is this blocking for the PR or just something to keep in mind?

libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -0,0 +1,321 @@
+//** Types for `withResource` */
```
michael-small, line 1: This meant to be moved down to the types?

michael-small (commented):
Immediate round of things I noticed, will do at least one more at a deeper level in a bit

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
rainerhahnekamp: No, it is really just a note for us. We will release earlier

michael-small:
As an rxResource enthusiast, I would like at least one test that has it, even though I know this is generic to any valid resource type. Even if it was just a cheap copy of an existing one, like this:

```ts
        it('can accept an rxResource', async () => {
          const UserStore = signalStore(
            { providedIn: 'root' },
            withState({ userId: 1 }),
            withResource((store) =>
              rxResource({
                params: store.userId,
                stream: ({ params: id }) => of(id + 1),
              }),
            ),
          );

          const userStore = TestBed.inject(UserStore);

          await wait();
          expect(userStore.value()).toBe(2);
        });
```

michael-small:
Oh, I pulled down your fork and tried to push up that commit fb77141 and assumed it would just open a PR to your fork. Oops, I didn't realize I could write directly to your fork's branch. Hope that change is good, I'll be careful next time knowing that.

michael-small (changes requested):
Two things and it's good

- Verify that the change I accidentally committed directly to this is good with you (`build` and `test` are fine)
- Please add something like that one `rxResource` test example somewhere, if not literally just that example. I suppose if you didn't mind I could just push it up too, apparently.

Otherwise, between this time and the times I have looked at the draft PR in the other repo, looks good to me.

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
rainerhahnekamp: Closing that one since we are not waiting

michael-small (approved):

PR #219, opened 2025-08-31
feat(rx-mutation): Make rx-mutation standalone

manfredsteyer:


libs/ngrx-toolkit/src/lib/mutation/rx-mutation.spec.ts
```diff
@@ -0,0 +1,594 @@
...
+      getCounter: () => counter,
+      onSuccessCalls: () => onSuccessCalls,
+      onErrorCalls: () => onErrorCalls,
+      lastOnSuccessParams: () => lastOnSuccessParams,
+      lastOnErrorParams: () => lastOnErrorParams,
+    };
+  });
+}
+
+describe('rxMutation', () => {
```
rainerhahnekamp, line 110: Since we do a test refactoring, I'll leave it here as it is. I will make sure that we have a test coverage which covers both cases (yours and mine)

libs/ngrx-toolkit/src/lib/mutation/rx-mutation.ts
```diff
@@ -10,8 +10,8 @@ import {
   tap,
 } from 'rxjs';
 
-import { concatOp, FlatteningOperator } from './flattening-operator';
-import { Mutation, MutationResult, MutationStatus } from './with-mutations';
+import { concatOp, FlatteningOperator } from '../flattening-operator';
+import { Mutation, MutationResult, MutationStatus } from './mutation';
 
 export type Func<P, R> = (params: P) => R;
```
rainerhahnekamp, line 17: I can't add comment to `RxMutationOptions` below, but it would again full names for the types.

I think we briefly chatted about a possible confusion when using both parameters `operation` and `operator`. Could we come up with a better name for `operator`? `flattener`, `flatteningOperator`, etc...

libs/ngrx-toolkit/src/lib/mutation/rx-mutation.ts
```diff
@@ -35,46 +35,70 @@ export interface RxMutationOptions<P, R> {
  *
  * The `operation` is the only mandatory option.
  *
+ * The returned mutation can be called as an async function and returns a Promise.
+ * This promise informs about whether the mutation was successful, failed, or aborted
```
rainerhahnekamp, line 39: if we decide to go with fulfilled, we wouldn't call it successful here.

libs/ngrx-toolkit/src/lib/mutation/rx-mutation.ts
```diff
@@ -35,46 +35,70 @@ export interface RxMutationOptions<P, R> {
...
+ *
+ * protected error = this.increment.error;
+ * protected isPending = this.increment.isPending;
+ * protected status = this.increment.status;
+ * protected value = this.increment.value;
+ * protected hasValue = this.increment.hasValue;
+ *
+ * async incrementCounter() {
+ *     const result = await this.increment({ value: 1 });
+ *     if (result.status === 'success') {
```
rainerhahnekamp, line 70: be aware in case we go with `fulfilled`.

libs/ngrx-toolkit/src/lib/mutation/rx-mutation.ts
```diff
@@ -35,46 +35,70 @@ export interface RxMutationOptions<P, R> {
...
+ *
+ * The mutation also provides several Signals such as error, status or isPending (see below).
+ *
+ * Example usage without Store:
+ *
  * ```typescript
- * export type Params = {
- *   value: number;
- * };
+ * private counterSignal = signal(0);
```
rainerhahnekamp, line 47: The example reads like it is a class. Should we add the class to it or just remove the access modifier and go with simple `const`?

So:

```typescript
const counterSignal = signal(0);

const increment = rxMutation({
    // ...
```

libs/ngrx-toolkit/src/lib/mutation/rx-mutation.ts
```diff
@@ -35,46 +35,70 @@ export interface RxMutationOptions<P, R> {
...
+ * }
  *
  * function calcSum(a: number, b: number): Observable<number> {
- *   return of(a + b);
+ *   return of(result).pipe(delay(500));
  * }
  * ```
  *
  * @param options
  * @returns
```
rainerhahnekamp, line 87: Maybe add here some additional text.


```suggestion
 * @returns the actual mutation function along tracking data as properties/methods
```

libs/ngrx-toolkit/src/lib/mutation/mutation.ts
```diff
@@ -0,0 +1,26 @@
+import { Signal } from '@angular/core';
+
+export type MutationResult<T> =
+  | {
+      status: 'success';
+      value: T;
```
rainerhahnekamp, line 6:
```suggestion
      value: Value;
```

libs/ngrx-toolkit/src/lib/mutation/mutation.ts
```diff
@@ -0,0 +1,26 @@
+import { Signal } from '@angular/core';
+
+export type MutationResult<T> =
```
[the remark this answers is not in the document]
manfredsteyer: Done

libs/ngrx-toolkit/src/lib/mutation/mutation.ts
```diff
@@ -0,0 +1,26 @@
...
+  | {
+      status: 'success';
```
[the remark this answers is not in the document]
manfredsteyer: Done

PR #226, opened 2025-09-02
fix: devtools-syncer missing case for extension

### LINKED ISSUE #213: withDevTools() not working in demo app?
Hello everyone, and thank you for your great work. I have a small problem with the DevTools integration. I wanted to add it to my app but was unsuccessful. The Redux DevTools screen just remains blank.

I just checked out the master branch of the NgRx Toolkit and started the demo app. I'm seeing the same behavior there (see screenshot).

Before taking the screenshot, I had selected “Read a Book” and “Write a Journal” and then deselected “Read a Book.”

Is there anything else I need to consider when using withDevTools()?

I am using macOS and the latest version of Chrome (version 139.0.7258.128).

Thank you very much!

Image

jdegand:
Closes #213

With these changes, you no longer get a blank screen when using Redux DevTools extension.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts
@@ -11,62 +11,60 @@ const dummyConnection: Connection = {
   send: () => void true,
 };
 
-/**
- * A service provided by the root injector is
- * required because the synchronization runs
- * globally.
- *
- * The SignalStore could be provided in a component.
- * If the effect starts in the injection
- * context of the SignalStore, the complete sync
- * process would shut down once the component gets
- * destroyed.
- */
 @Injectable({ providedIn: 'root' })
 export class DevtoolsSyncer implements OnDestroy {
-  /**
-   * Stores all SignalStores that are connected to the
-   * DevTools along their options, names and id.
-   */
   #stores: StoreRegistry = {};
   readonly #isBrowser = isPlatformBrowser(inject(PLATFORM_ID));
-  readonly #trackers = [] as Tracker[];
+  readonly #trackers: Tracker[] = [];
   readonly #devtoolsConfig = {
     name: 'NgRx SignalStore',
     ...inject(REDUX_DEVTOOLS_CONFIG, { optional: true }),
   };
 
-  /**
-   * Maintains the current states of all stores to avoid conflicts
-   * between glitch-free and glitched trackers when used simultaneously.
-   *
-   * The challenge lies in ensuring that glitched trackers do not
-   * interfere with the synchronization process of glitch-free trackers.
-   * Specifically, glitched trackers could cause the synchronization to
-   * read the current state of stores managed by glitch-free trackers.
-   *
-   * Therefore, the synchronization process doesn't read the state from
-   * each store, but relies on #currentState.
-   *
-   * Please note, that here the key is the name and not the id.
-   */
   #currentState: Record<string, object> = {};
   #currentId = 1;
 
   readonly #connection: Connection = this.#isBrowser
-    ? window.__REDUX_DEVTOOLS_EXTENSION__
-      ? window.__REDUX_DEVTOOLS_EXTENSION__.connect(this.#devtoolsConfig)
-      : dummyConnection
+    ? this.#initDevtoolsConnection()
     : dummyConnection;
 
   constructor() {
     if (!this.#isBrowser) {
+      console.warn(
+        '[DevtoolsSyncer] Not running in browser. DevTools disabled.',
+      );
       return;
     }
   }
 
   ngOnDestroy(): void {
     currentActionNames.clear();
+  }
+
+  #initDevtoolsConnection(): Connection {
+    const extension = window.__REDUX_DEVTOOLS_EXTENSION__;
+    if (!extension) {
+      console.warn('[DevtoolsSyncer] Redux DevTools extension not found.');
+      return dummyConnection;
+    }
+
+    try {
+      if (typeof extension.connect === 'function') {
+        return extension.connect(this.#devtoolsConfig);
+      } else {
+        console.warn(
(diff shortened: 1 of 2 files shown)


jdegand:
On Line 26 in `with-devtools`,  EXISTING_NAMES is unused.  I left it alone (I can't comment on that line in a code review).

I removed the unneeded `id` variable in `with-devtools` in this PR.  I noticed this during the course of my investigation of the error.

rainerhahnekamp:
I am not Michael, but I took a quick look. Would it be possible to construct a test case that causes the bug?

michael-small:
> On Line 26 in `with-devtools`, EXISTING_NAMES is unused. I left it alone (I can't comment on that line in a code review).

Sounds good to remove for me

> I removed the unneeded `id` variable in `with-devtools` in this PR. I noticed this during the course of my investigation of the error.

Good catch

michael-small (approved):
Thank you for getting to this as I had not yet. This is cleaner than what I was considering doing. 

I think it looks good. Approved on the assumption that a test case comes along.

libs/ngrx-toolkit/src/lib/devtools/tests/devtools-syncer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    (window as any).__REDUX_DEVTOOLS_EXTENSION__ = { connect: connectSpy };
+
+    TestBed.resetTestingModule();
+    TestBed.configureTestingModule({
+      providers: [{ provide: PLATFORM_ID, useValue: 'browser' }],
+    });
+  });
+
+  function createTrackerMock(): Tracker {
+    const onChangeMock = jest.fn();
```
jdegand, line 22: Didn't think about how this will translate to Vitest, but it should be fine.

libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts
```diff
@@ -8,7 +8,7 @@ import { DevtoolsInnerOptions } from './devtools-feature';
 import { Connection, StoreRegistry, Tracker } from './models';
 
 const dummyConnection: Connection = {
-  send: () => void true,
+  send: () => true,
 };
```
jdegand, line 13: Added a separate test file.  I did look at other tests and noticed deprecated use of `TestBed.flushEffects()` should be able to quickly replaced with `TestBed.tick()`.  Best to fix during Vitest conversion PR.

libs/ngrx-toolkit/src/lib/devtools/tests/devtools-syncer.spec.ts
jdegand: Something like this: 

```ts
(tracker.onChange as ReturnType<typeof vi.fn>).mockImplementation((cb) => {
  cb({ [id]: { count: 42 } });
});
```

libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts
```diff
@@ -54,21 +54,47 @@ export class DevtoolsSyncer implements OnDestroy {
...
   }
 
   ngOnDestroy(): void {
     currentActionNames.clear();
   }
 
+  #initDevtoolsConnection(): Connection {
+    const extension = window.__REDUX_DEVTOOLS_EXTENSION__;
+    if (!extension) {
+      console.warn('[DevtoolsSyncer] Redux DevTools extension not found.');
```
rainerhahnekamp, line 75: i think it can be annoying to throw warnings if DevTools are not
there. Especially since they can also be enabled in production.

libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts
```diff
@@ -54,21 +54,47 @@ export class DevtoolsSyncer implements OnDestroy {
...
-    ? window.__REDUX_DEVTOOLS_EXTENSION__
-      ? window.__REDUX_DEVTOOLS_EXTENSION__.connect(this.#devtoolsConfig)
-      : dummyConnection
+    ? this.#initDevtoolsConnection()
     : dummyConnection;
 
   constructor() {
     if (!this.#isBrowser) {
-      return;
+      console.warn(
```
rainerhahnekamp, line 62: No warnings if we are on the server. We had
this before and users were complaining that
we are filling up their logs.

libs/ngrx-toolkit/src/lib/devtools/tests/devtools-syncer.spec.ts
```diff
@@ -0,0 +1,96 @@
+import { PLATFORM_ID } from '@angular/core';
```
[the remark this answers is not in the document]
rainerhahnekamp: I ran all of these tests against the current main branch.  
All of them passed except *"should handle extension absence gracefully"*.  
That test failed only because it checks whether a warning message is shown when
extensions aren’t available.

Issue #213, however, shows an error from `renameStore`, which points to a
different problem than the one addressed in this PR.

We need a test that reproduces the reported error.

Other than that, I do think that you don't have to create a separate file.
Adding it to `connecting.spec.ts` should work.

rainerhahnekamp (changes requested):
Sorry to spoil the party, but there are a few things that need attention:

- The actual bug doesn’t seem to be fixed. I left a comment in the tests, but
  instead of addressing that, you added extra warning messages (which had been
  removed on purpose before). Also, there is still no test that reproduces the
  reported issue.
- I generally appreciate PRs that include refactorings. The problem here is
  that by applying a different style, I also have to review those changes,
  which shifts the focus away from the bug fix. If you want to do refactorings,
  please open them in a separate PR.

Once we have a test that reproduces the bug (and fixes it), I think we’ll be on the right
track.

jdegand:
Closing in favor of #234.

PR #234, opened 2025-09-06
fix(devtools): prevent rename race on unmount and correct state cleanup  #213

wolfmanfx:
Closes [#213](https://github.com/angular-architects/ngrx-toolkit/issues/213)

### Summary
- Prevents runtime errors and blank DevTools by making rename operations stable, idempotent, and safe during component unmount.
- Corrects state cleanup to avoid corrupting DevTools state when a store is removed.

### Changes
- **Rename stability**
  - Added `renameStoreById(id, newName)` and switched `withDevtools()` to call it.
  - Centralized shared logic in private `#applyRename(id, oldName, newName)`.
  - Made rename idempotent (no-op if name unchanged).
  - Duplicate-name checks ignore the current store.

- **State cleanup fix**
  - In `removeStore`, preserve entries using the iterator key (`storeName`) instead of the removed store’s `name`.

- **Demo consistency**
  - `TodoStore.remove` now also removes the id from `selectedIds` to keep selection and entities in sync. This leaded also to errors.

### Why
- Unchecking/destroying a detail component could trigger a rename after its store was removed; name-based lookup returned null/undefined and threw.
- Rebuilding `#currentState` using the removed store’s `name` re-keyed all survivors incorrectly, corrupting the DevTools view.

### How to verify
1. Start the demo and open Redux DevTools.
2. Select multiple todos to render details; toggle selection (check/uncheck).
3. Delete a selected todo.
4. Expected:
   - No runtime errors.
   - DevTools shows stable, correctly renamed store entries.
   - No state corruption when removing/renaming stores.

### Affected files
- `libs/ngrx-toolkit/src/lib/devtools/internal/devtools-syncer.service.ts`
- `libs/ngrx-toolkit/src/lib/devtools/with-devtools.ts`
- `apps/demo/src/app/devtools/todo-store.ts`

### Notes
- No breaking changes expected.
- Builds and demo run pass locally.

### Commands
```bash
pnpm build:all
pnpm start
pnpm test:e2e
```

### DIFF
--- a/apps/demo/src/app/devtools/todo-store.ts
+++ b/apps/demo/src/app/devtools/todo-store.ts
@@ -30,7 +30,14 @@ export const TodoStore = signalStore(
       },
 
       remove(id: number) {
-        updateState(store, 'remove todo', removeEntity(id));
+        updateState(
+          store,
+          'remove todo',
+          removeEntity(id),
+          ({ selectedIds }) => ({
+            selectedIds: selectedIds.filter((selectedId) => selectedId !== id),
+          }),
+        );
       },
 
       toggleFinished(id: number): void {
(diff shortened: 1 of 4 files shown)


jdegand:
In `with-devtools`, on line 26, you can remove `EXISTING_NAMES`, the import of `InjectionToken` and you can safely remove `const id = String(store[uniqueDevtoolsId]());` inside `onInit`.

PR #240, opened 2025-09-13
chore(app): restructure mutations for better typing/naming

michael-small:
As I looked over the app examples for writing the documentation, I noticed two things that could be improved for clarity now that best practices have been hashed out more.

- Typing: Rather than explicit types in the function's generic params for `httpMutation`, it is better to be typed with: 
    - the parameter type in the actual `request` parameters
    - response in the response callbacks and/or `parse`.
- Naming: the example for the standalone functions for mutations was made with `rxMutation` in mind, but ended up encompassing `httpMutation` too


rainerhahnekamp:
@michael-small sorry that i didn't review it. Could please update from main and check if this is still valid?

michael-small:
@rainerhahnekamp Rebased and ready for another reveiw

apps/demo/src/app/counter-mutation-functions/counter-mutation-functions.ts
```diff
@@ -41,13 +38,13 @@ export class CounterRxMutation {
...
-  private saveToServer = httpMutation<Params, CounterResponse>({
-    request: (p) => ({
+  private saveToServer = httpMutation({
+    request: (p: Params) => ({
       url: `https://httpbin.org/post`,
       method: 'POST',
       body: { counter: p.value },
-      headers: { 'Content-Type': 'application/json' },
     }),
+    parse: (response) => response as CounterResponse,
```
rainerhahnekamp, line 47: Shouldn't we use here Zod?

michael-small: I have been following the precedent from discussions around when I made the docs. Which was mostly me following examples from the RFC and the tests/playground app. Doing `parse: (response) => response as XYZ` seemed to be the precedent, so I haven't considered that validation libraries like Zod was possible until now.

If we want to use Zod, then I can pull it into the playground's dependencies and see how it works with `parse`.

Issue #178, opened 2025-06-04
bug: DevtoolsFeature is not exported


HansKrywalsky:
# Description

Hey there, we have a wrapper function around adding `withDevtools`, to replace it with `withDevtoolsStub` on production for example.

I want to pass `DevtoolsFeature[]` into this wrapper function, but i can't use `DevtoolsFeature` because it is not exported i guess? (using `moduleResolution: 'bundler'`

michael-small:
@rainerhahnekamp I believe that exporting the type would be reasonable. Any reasons to the contrary?

rainerhahnekamp:
No, please export it

PR #251, opened 2025-10-07
chore: export `DevtoolsFeature` type

michael-small:
Would close out https://github.com/angular-architects/ngrx-toolkit/issues/178


libs/ngrx-toolkit/src/lib/devtools/with-devtools.ts
```diff
@@ -8,12 +8,15 @@ import {
 } from '@ngrx/signals';
 import { DefaultTracker } from './internal/default-tracker';
 import {
-  DevtoolsFeature,
+  DevtoolsFeature as DevtoolsFeatureInternal,
```
rainerhahnekamp, line 11: Is there a reason why you rename it here and reverse it later?

rainerhahnekamp (changes requested):

Issue #231, opened 2025-09-05
@ngrx/signals/events: How to display event name/type in Redux DevTools? Event name always: 'Store Update'


DmitriyIvanko:
**How to display the name for events (@ngrx/signals/events) in Redux DevTools?**

To display the state of signalStore in ReduxDevTools, you need to use:

```
export const ServerStore = signalStore(
  withDevtools('book'),
  withEntities({ entity: type<Book>() }),
  withState<State>({ isLoading: false }),
  // some other signal store settings;
)

```
To display the name of the some method, just use updateState instead of patchState:

```
withMethods((store) => ({
  onLoadingChanged(isLoading: boolean): void {
    updateState(store, 'onLoadingChanged', () => ({ isLoading }));
  },
})),
```

But how to display the name for events (@ngrx/signals/events)?:

```
withReducer(
  on(bookApiEvents.get, () => ({ isLoading: true })),
),
```

Here there is no way to apply updateState, and such events are always shown in Redux DevTools as 'Store Update' which makes debugging very difficult.

I try this:
```

on(bookApiEvents.getSuccess, (event, state: State) => {
  updateState<State>(state as unknown as WritableStateSource<State>, event.type, () => ({ isLoading: false }));

  return (): State => ({ isLoading: false });
}),
```

But there is console error (can not use this)

Please help, thank you!

jdegand:
You cannot add `event.type` to `withReducer`. 

You can try:

```
withMethods((store) => ({
  handleGetSuccess(event: BookApiEvent): void {
    updateState(store, event.type, () => ({ isLoading: false }));
  },
}))
```

Dispatch `handleGetSuccess` manually when the event occurs.

I think code modification and refactoring may be necessary to enable this functionality.

rainerhahnekamp:
@DmitriyIvanko, yes, that's the right place 😅.

So here's the situation: as you know, the DevTools integration allows you to define an action’s name only via `updateState`. The problem with events is that they internally call `patchState`, which leaves us without a way to define the action name.

The only solution I see is similar to how the global store works: catch all events and use the event name as the action name.

@wolfmanfx showed in [this article](https://www.angulararchitects.io/blog/the-new-event-api-in-ngrx-signal-store/) a possible approach using meta reducers. The challenge is that this might collide with the default DevTools behavior that watches the state, so we’d need to figure out how to resolve that. I am also not sure if these meta reducers guarantee that we will be executed after all reducers were executed.

[...]

PR #266, opened 2025-12-26
Alternative Event Tracking: `withTrackedReducer`

rainerhahnekamp:
## Summary

This PR introduces `withTrackedReducer` as an alternative approach to tracking reducer-based state changes in Redux DevTools. It addresses limitations in the original `withEventsTracking` implementation.

The original implementation (based on #231) attempted to automatically track reducer-based state changes via the glitch tracker. However, this approach has fundamental limitations that cannot be resolved with the specified approach.

Several edge cases are documented in the skipped tests within `with-events-tracking.spec.ts`.

This PR introduces an alternative way, where end users have to use the wrapperd function `withTrackedReducer` which calls internally `updateState` using the event's name.

If we proceed with this approach, the following items need to be completed:

- [ ] Update documentation to explain `withTrackedReducer` usage and when to use it vs `withReducer`
- [ ] Update demo applications to showcase the new API
- [ ] Remove or deprecate code from the initial `withEventsTracking` approach (if applicable)
- [ ] Add validation/runtime checks to ensure `withTrackedReducer` is only used with stores that have devtools and glitch tracking enabled

### DIFF
--- a/libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
@@ -6,6 +6,7 @@ import {
   SignalStoreFeature,
   signalStoreFeature,
   type,
+  withHooks,
 } from '@ngrx/signals';
 import {
   EventCreator,
@@ -15,22 +16,14 @@ import {
 import { tap } from 'rxjs/operators';
 import { GLITCH_TRACKING_FEATURE } from './features/with-glitch-tracking';
 import { updateState } from './update-state';
-import { DEVTOOL_PROP } from './with-devtools';
+import { DEVTOOL_FEATURE_NAMES } from './with-devtools';
 
-export function withTrackedReducer<
-  State extends object,
-  DevtoolsFeatureName extends string,
->(
+export function withTrackedReducer<State extends object>(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ...caseReducers: CaseReducerResult<State, any>[]
 ): SignalStoreFeature<
   EmptyFeatureResult & {
     state: State;
-    props: {
-      [DEVTOOL_PROP]: typeof GLITCH_TRACKING_FEATURE extends DevtoolsFeatureName
-        ? DevtoolsFeatureName
-        : 'NO GLITCHED TRACKING ACTIVATED';
-    };
   },
   EmptyFeatureResult
 > {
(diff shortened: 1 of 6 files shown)


apps/demo/src/app/events-sample/events-sample.component.ts
```diff
@@ -0,0 +1,196 @@
...
+          <input
+            matInput
+            [(ngModel)]="filterText"
+            (ngModelChange)="dispatch.filterUpdated({ filter: $event })"
+            placeholder="Filter by title or author..."
+          />
+          <mat-icon matSuffix>search</mat-icon>
+        </mat-form-field>
+
+        <button mat-raised-button color="primary" (click)="addRandomBook()">
```
michael-small: nit: though the demo is lowkey, I would like things like requiring a `type` on `button` elements like this. I am going about backfilling this for existing demo code.

apps/demo/src/app/events-sample/events-sample.component.ts
```diff
@@ -0,0 +1,196 @@
...
+import { MatIconModule } from '@angular/material/icon';
+import { MatInputModule } from '@angular/material/input';
+import { MatToolbarModule } from '@angular/material/toolbar';
+import { injectDispatch } from '@ngrx/signals/events';
+import { bookEvents } from './book-events';
+import { BookStore } from './book.store';
+
+@Component({
+  selector: 'demo-events-sample',
+  standalone: true,
```
michael-small:
```suggestion
```

rainerhahnekamp:
@michael-small @wolfmanfx 

I see it as temporarily. In the end, we need support for meta-reducers in the core...

libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+  SignalStoreFeature,
+  signalStoreFeature,
+  type,
+} from '@ngrx/signals';
+import { ReducerEvents, withEventHandlers } from '@ngrx/signals/events';
+import { tap } from 'rxjs/operators';
+import { updateState } from './update-state';
+
+export function withTrackedReducer<State extends object>(
+  ...caseReducers: CaseReducerResult<State, any>[]
```
rainerhahnekamp: For some reason, I had to use `any` here. The `EventCreator` - as it is in the original version - didn't work. I guess that this is due to different settings in the `tsconfig.json`. 

Unless, we won't get meta reducers, I wouldn't investigate this any further. The only thing I would change is to replace `any` with `unknown`.

wolfmanfx (approved):

Issue #242, opened 2025-09-19
withResources with fail resource is blocking the store.


luandaja:
When a resource is in a failing state for any reason and you try to `patchState`, the console will display an error and the action will be skipped.

<img width="1770" height="562" alt="Image" src="https://github.com/user-attachments/assets/a4e5ba73-a33e-4b45-a85e-d2a6e98102b8" />

This behavior can be reproduced with this store and calling any increment method.

If I'm doing something wrong, please let me know.

```
import { withResource } from '@angular-architects/ngrx-toolkit';
import { resource } from '@angular/core';
import { patchState, signalStore, withMethods, withState } from '@ngrx/signals';

export const Store = signalStore(
  {
    providedIn: 'root',
  },
  withState({
    count: 0,
  }),

  withResource(() => ({
    test: resource({
      loader: () => {
        console.log('test');
        // throw new Error('test');
        return Promise.reject('test');
      },
    }),
    // test2: httpResource(() => 'https://www.not-exist'),
  })),
  withMethods((store) => ({
    increment() {
      patchState(store, { count: store.count() + 1 });
    },
    increment2() {
      patchState(store, (state) => ({ ...state, count: state.count + 1 }));
    },
  }))
);

```

Checking the `ngrx/platform` `patchState` code, the issue is because when we call `patchState`, this call internally calls getState that gets the value. 

<img width="1378" height="450" alt="Image" src="https://github.com/user-attachments/assets/81ca7237-0f64-4968-bdeb-e59214ee80dc" />

luandaja:
In the `withResources` code, I moved the Value from state to props and now it is working, but I am not sure about that change (i'll research more).

However, that confirms the issue. When the resource is in a fail state, it cannot access the value, and if we keep the value as part of the `state`, we will call that signal as part of the ngrx/signals `patchState`.

<img width="1308" height="790" alt="Image" src="https://github.com/user-attachments/assets/4b89546d-076d-45ec-b619-46089a2d394c" />

rainerhahnekamp:
@luandaja . That’s the behavior of angular resource. Accessing a value if it is in error state, it will always throw

luandaja:
> [@luandaja](https://github.com/luandaja) . That’s the behavior of angular resource. Accessing a value if it is in error state, it will always throw

[...]

PR #267, opened 2025-12-26
Non-Failing `withResource`

rainerhahnekamp:
Addresses #242


rainerhahnekamp:
@michael-small 

> I'll give a formal review when this is marked ready, but this looks like a solid plan.

I see it as temporarily. In the end, we need support for meta-reducers in the core...

michael-small:
Would it be possible to set a provider for an app config for the default error behavior?

rainerhahnekamp:
> Would it be possible to set a provider for an app config for the default error behavior?

Good idea @michael-small

michael-small:
Reason: With our mindset and approach for error handling having shifted to use a proxy, we have closed this in favor of https://github.com/angular-architects/ngrx-toolkit/pull/270

PR #295, opened 2026-04-03
feat(devtools): display event payloads for withTrackedReducer

### LINKED ISSUE #294: RFC: Expose Event Payload in Redux DevTools via `withTrackedReducer`
# RFC: Expose Event Payload in Redux DevTools via `withTrackedReducer`

## 1. Summary

When using `withTrackedReducer`, the Redux DevTools extension currently logs only the string name of a dispatched event (e.g., `[User] Update Age`). The associated payload is silently discarded, leaving developers without visibility into what data drove a state transition.

This RFC proposes three coordinated, non-breaking internal changes to surface the full event object — including its payload — in the DevTools action log:

1. Refactor `withTrackedReducer` to forward the full event object instead of extracting only `event.type`.
2. Add an overload to the internal `updateState` utility that accepts an `EventInstance` (any object with a `type: string` property) as its action descriptor.
3. Update `DevtoolsSyncer` to pass that full object to the Redux DevTools extension, making payload data visible alongside the action type.

---

### 2.1 Developer Experience Gap

NgRx Signal Store's `withTrackedReducer` is designed to bring reducer-style determinism to the Signal Store. However, the current DevTools integration surfaces only half of the picture: the *name* of what happened, but not *what* happened. In practice, this forces developers to fall back to `console.log` statements inside their reducers to inspect the payload driving each transition — exactly the kind of manual introspection that DevTools is meant to eliminate.

### 2.2 Parity with Traditional NgRx

In a classic NgRx `Store` setup backed by `@ngrx/store-devtools`, every dispatched action appears in the DevTools panel with its full serialized payload. A developer can click any entry in the action log and immediately inspect the data. `withTrackedReducer` should provide the same level of observability; the current behaviour is a silent regression in DX for teams migrating from, or running alongside, the traditional NgRx store.

---

## 3. Design

**3.1 — `withTrackedReducer`: pass the full event, not just the type**
```typescript
// Before
updateState(store, event.type, nextState);

// After
updateState(store, event, nextState);
```

**3.2 — `updateState`: add an object overload**
```typescript
// Existing overload — unchanged
export function updateState<State extends object>(
  stateSource: WritableStateSource<State>,
  action: string,
  ...updaters: Array<
    Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
  >
): void;

// New overload
export function updateState<State extends object>(
  stateSource: WritableStateSource<State>,
  eventInstance: EventInstance<string, any>,
  ...updaters: Array<
    Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
  >
): void;

// Implementation normalises both shapes
export function updateState<State extends object>(
  stateSource: WritableStateSource<State>,
  eventOrEventType: string | EventInstance<string, any>,
  ...updaters: Array<
    Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
  >
): void {
  if (typeof eventOrEventType === 'string') {
    currentActionNames.add(eventOrEventType);
  } else {
    currentEvents.add(eventOrEventType);
  }
  return originalPatchState(stateSource, ...updaters);
}
```

**3.3 — `DevtoolsSyncer`: forward the object to the extension**
```typescript
// Before
this.#connection.send({ type }, this.#currentState);

// After — Redux DevTools natively supports action objects
this.#connection.send({ type, payload }, this.#currentState);
```

**Result in DevTools:**

```code

// Before   
       
▶ [User] Update Age

type: "[User] Update Age"

// After

▶ [User] Update Age

type: "[User] Update Age"
payload: {
    userId: "u_4821"
    age: 31
}

```

---

## 4. Adoption Strategy

Strictly non-breaking. The `string` overload is preserved. No consumer code changes. No migration guide needed. Safe to ship as a **minor** release.

---

## 5. Implementation Plan

I have a working local fork with all three changes verified against the Chrome DevTools extension. The PR will include unit tests for both `updateState` overloads and E2E verification that the payload appears correctly in the action log.

OleksandrOleniuk:
Closes [#294 RFC](https://github.com/angular-architects/ngrx-toolkit/issues/294)

This PR implements the enhancements proposed in the RFC, upgrading the DevTools integration to display full event payloads when using withTrackedReducer.

### DIFF
--- a/libs/ngrx-toolkit/src/lib/devtools/internal/current-action-names.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/internal/current-action-names.ts
@@ -1,1 +1,5 @@
+import { EventInstance } from '@ngrx/signals/events';
+
 export const currentActionNames = new Set<string>();
+
+export const currentEvents = new Set<EventInstance<string, any>>();
--- a/apps/demo/e2e/events-sample.spec.ts
+++ b/apps/demo/e2e/events-sample.spec.ts
@@ -1,0 +1,63 @@
+import { Action } from '@ngrx/store';
+import { expect, test } from '@playwright/test';
+
+test.describe('events-sample + devtools', () => {
+  test('DevTools are syncing events and payloads', async ({ page }) => {
+    await page.goto('');
+
+    await page.evaluate(() => {
+      window['devtoolsSpy'] = [];
+
+      window['__REDUX_DEVTOOLS_EXTENSION__'] = {
+        connect: () => {
+          return {
+            send: (data: Action) => {
+              window['devtoolsSpy'].push(data);
+            },
+          };
+        },
+      };
+    });
+
+    await page.getByRole('link', { name: 'Events + DevTools Sample' }).click();
+
+    // Filter by title
+    await page.getByPlaceholder('Filter by title or author...').fill('Hobbit');
+
+    // Add Book (Random)
+    await page.getByRole('button', { name: 'Add Book' }).click();
+
+    // Clear Selection
+    await page.getByRole('button', { name: 'Clear Selection' }).click();
+
+    const devtoolsActions = await page.evaluate(() => window['devtoolsSpy']);
+
+    // Check if filterUpdated event was tracked with the expected payload
+    const filterEvent = devtoolsActions.find(
+      (a) => a.type === '[Book Store] filterUpdated',
+    );
+    expect(filterEvent).toBeDefined();
+    expect(filterEvent).toEqual({
+      type: '[Book Store] filterUpdated',
+      payload: { filter: 'Hobbit' },
+    });
+
+    // Check selectionCleared event
+    const clearSelectionEvent = devtoolsActions.find(
+      (a) => a.type === '[Book Store] selectionCleared',
+    );
+    expect(clearSelectionEvent).toBeDefined();
+    expect(clearSelectionEvent).toEqual({
+      type: '[Book Store] selectionCleared',
+    });
+
+    // Check bookAdded event (payload will contain random book, so just check if payload has book object)
+    const bookAddedEvent = devtoolsActions.find(
+      (a) => a.type === '[Book Store] bookAdded',
+    );
+    expect(bookAddedEvent).toBeDefined();
+    expect(bookAddedEvent.payload).toBeDefined();
+    expect(bookAddedEvent.payload.book).toBeDefined();
+    expect(bookAddedEvent.payload.book).toHaveProperty('title');
+  });
+});
--- a/docs/docs/with-devtools.md
+++ b/docs/docs/with-devtools.md
@@ -50,6 +50,10 @@ patchState(this.store, { loading: false });
 
 // updateState is a wrapper around patchState and has an action name as second parameter
 updateState(this.store, 'update loading', { loading: false });
+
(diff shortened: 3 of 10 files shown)


libs/ngrx-toolkit/src/lib/devtools/internal/current-action-names.ts
```diff
@@ -1 +1,5 @@
+import { EventInstance } from '@ngrx/signals/events';
+
 export const currentActionNames = new Set<string>();
+
+export const currentEvents = new Set<EventInstance<string, any>>();
```
rainerhahnekamp, line 5: to be honest, I think we don't even have to bring a queue for the events. `withTrackedReducer` should consume the event and immediately send it devtools.

The action names were only necessary when are collecting the state changes via an `effect`.

Because of that, I think this whole PR could be much shorter.

rainerhahnekamp (changes requested):
Thanks for the quick PR. Please find my comment below

libs/ngrx-toolkit/src/lib/devtools/internal/current-action-names.ts
OleksandrOleniuk: Thanks for the comment

I just based it on the existing functionality. Could you explain what do you mean by 

> withTrackedReducer should consume the event and immediately send it devtools

I could inject `DevtoolsSyncer` after `updateState` and call some syncing, but i don't think this is a nice way

```typescript

caseReducers.map((caseReducer) =>
  events.on(...caseReducer.events).pipe(
    tap((event) => {
      const state = untracked(() => getState(store));
      const result = caseReducer.reducer(event, state);
      const updaters = Array.isArray(result) ? result : [result];

      updateState(store, event.type, ...updaters);
    }),
  ),
),

```

rainerhahnekamp: The only necessary change to the devtools was to use `event.type` in `with-tracked-reducer` and extend `updateState`. I don't see the need for the `currentEvent`, respectively I don't understand the additional benefit.

sqrter: Hey @rainerhahnekamp, could you please take a look at #302 If I got your comments right, it would be nice to have it merged and make #294 completed. Thanks!

rainerhahnekamp: @sqrter, this one here is almost done?

@OleksandrOleniuk, if you don't have time to finish it, can @sqrter take over?

OleksandrOleniuk: @rainerhahnekamp , I think you can continue with #302 , since I didn't manage to get to the point with this one

sqrter: Hey @rainerhahnekamp, so do you mind if we continue with the #302 then? I'd be happy to adjust it if needed, if you could reopen it. Thanks

rainerhahnekamp: @sqrter yes we can continue there.

@OleksandrOleniuk, thanks for your help so far

Issue #292, opened 2026-03-07
RFC: resource snapshots and `withResource`


michael-small:
How do we handle the new resource snapshot API in conjunction with `withResource`?

> You can create new resources from snapshots using [resourceFromSnapshots](https://angular.dev/api/core/resourceFromSnapshots). This enables composition with signal APIs like [computed](https://angular.dev/api/core/computed) and [linkedSignal](https://angular.dev/api/core/linkedSignal) to transform resource behavior.

- About snapshots
- Snapshot composition limitations
- Open question on how to integrate resource snapshots + `withResource`

Basic reference Stackblitz I made for type inference, nothing toolkit related itself: https://stackblitz.com/edit/stackblitz-starters-kohef1nl?file=src%2Fmain.ts

## Resource snapshots: "enables composition with signal APIs ... to transform resource behavior"

21.2.0 introduced [resource composition with snapshots](https://angular.dev/guide/signals/resource#resource-composition-with-snapshots).

> - A [ResourceSnapshot](https://angular.dev/api/core/ResourceSnapshot) is a structured representation of a resource's current state. Every resource has a snapshot property that provides a signal of its current state.
> - Each snapshot contains a `status` and either a `value` or an `error`.
> - You can create new resources from snapshots using [resourceFromSnapshots](https://angular.dev/api/core/resourceFromSnapshots). This enables composition with signal APIs like [computed](https://angular.dev/api/core/computed) and [linkedSignal](https://angular.dev/api/core/linkedSignal) to transform resource behavior.

Example from the docs:

```ts
import {linkedSignal, resourceFromSnapshots, Resource, ResourceSnapshot} from '@angular/core';

function withPreviousValue<T>(input: Resource<T>): Resource<T> {
  const derived = linkedSignal<ResourceSnapshot<T>, ResourceSnapshot<T>>({
    source: input.snapshot,
    computation: (snap, previous) => {
      if (snap.status === 'loading' && previous && previous.value.status !== 'error') {
        // When the input resource enters loading state, we keep the value
        // from its previous state, if any.
        return {status: 'loading' as const, value: previous.value.value};
      }
      // Otherwise we simply forward the state of the input resource.
      return snap;
    },
  });
  return resourceFromSnapshots(derived);
}

@Component({
  /*... */
})
export class AwesomeProfile {
  userId = input.required<number>();
  user = withPreviousValue(httpResource(() => `/user/${this.userId()}`));
[...]

PR #298, opened 2026-04-20
feat: widen `withResource` to accept `Resource`

michael-small:
https://github.com/angular-architects/ngrx-toolkit/issues/292

This will allow using resource snapshots

This is based on top of updating to Angular 21.2.x since that supports snapshots.


rainerhahnekamp (commented):
@michael-small, can we have a separate PR which only updates Angular to 21.2?

michael-small:
@rainerhahnekamp done https://github.com/angular-architects/ngrx-toolkit/pull/299

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
```diff
@@ -538,6 +538,9 @@ describe('withResource', () => {
         IsEqual<typeof _store.digitValue, Signal<number | undefined>>
       >;
     });
+    it('only exposes reload methods for reloadable resources', () => {
+      // TODO - can type tests for _reload be done?
```
msmallest, line 542: Open question I am not sure of myself

rainerhahnekamp: good question. it would be something where you call `withResource` without the `signalStore`

libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -20,7 +21,33 @@ import {
   withProps,
 } from '@ngrx/signals';
 
+// TODO - add other specifics of HttpResourceRef (headers, statusCode, progress)
```
msmallest, line 24: This is what I was wondering in the original upgrade to 21.2 PR question I left recently. 

Should adding specific properties like that ^  be done in this PR or its own PR?

michael-small:
This is such a mess that I am restarting fresh on a new branch/PR lol

PR #299, opened 2026-04-21
chore: update to Angular 21.2

michael-small:


rainerhahnekamp (commented):
Hey Michael, I think that this is not just the update but includes also the snapshot?

Would it be possible only to touch those files that are really necessary for the upgrade to 21.2?

michael-small:
Ah, I forgot that I had done some preliminary stuff with the snapshots. The implementation of `withResource` did have to be changed a bit for the new shape of resources including the snapshot property so it could continue working. But these changes went beyond that like you saw. 

I think I can more cleanly just re-do this update fresh on a different branch.

Should I expose the `snapshot` property to be exposed to the store for a resource for parity with 21.2 resources now all having a `snapshot` property?

rainerhahnekamp:
> Should I expose the snapshot property to the store for a resource for parity with 21.2 resources, now all having a snapshot property?

So I'd say that all those demo files that got changed shouldn't be in the PR (unless there is a real reason for it).

Since 21.2 has the `snapshot` property, there is a chance that we are running into a compilation error with `withResource`. So if the compilation/build fails, you would have to update `withResource` as well. If it does not fail, I think, adding `snapshot` to `withResource` could be even a separate PR.

msmallest:
@rainerhahnekamp before I close this on my main account:

## Question 1: `snapshot` property, and where to add it

>  If it does not fail, I think, adding snapshot to withResource could be even a separate PR.

Since the update did not fail, I am wondering: should the `snapshot` property be exposed by

- It's own PR, right on top of the 21.2 bump that was just merged?
- OR At the same time as the PR to widen `withResource` to be able to take `Resource`?

## Question 2: `httpResource` exposed properties, when?

We mentioned at some point that exposing some `httpResource` specific properties should be done.

Should that be done

- In it's own PR
- OR At the same time as the PR to widen `withResource` to be able to take `Resource`?

michael-small:
Did this cleanly on my other GH account. Thank you me.

PR #301, opened 2026-04-24
feat: widen `withResource` to `Resource` beyond `ResourceRef`

michael-small:
Issue: https://github.com/angular-architects/ngrx-toolkit/issues/292

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-resource.spec.ts
+++ b/libs/ngrx-toolkit/src/lib/with-resource.spec.ts
@@ -269,6 +269,72 @@ describe('withResource', () => {
       });
 
       describe('extra properties checks', () => {
+        describe('extra properties go to `props` so they cannot be written to', () => {
+          it('for unnamed writables', async () => {
+            const Store = signalStore(
+              { providedIn: 'root', protectedState: false },
+              withResource(() => restResourceWritable(() => 'a')),
+            );
+
+            const store = TestBed.inject(Store);
+
+            await wait();
+
+            expect(store.stuff()).toBe('a stuff');
+
+            // @ts-expect-error - extra properties should not be patchable
+            patchState(store, { stuff: 'b stuff' });
+          });
+          it('for unnamed readables', async () => {
+            const Store = signalStore(
+              { providedIn: 'root', protectedState: false },
+              withResource(() => restResourceReadable(() => 'a')),
+            );
+
+            const store = TestBed.inject(Store);
+
+            await wait();
+
+            expect(store.stuff()).toBe('a stuff');
+
+            // @ts-expect-error - extra properties should not be patchable
+            patchState(store, { stuff: 'b stuff' });
+          });
+          it('for named readables', async () => {
+            const Store = signalStore(
+              { providedIn: 'root', protectedState: false },
+              withResource(() => ({
+                name: restResourceReadable(() => 'a'),
+              })),
+            );
+
+            const store = TestBed.inject(Store);
+
+            await wait();
+
+            expect(store.nameStuff()).toBe('a stuff');
+
+            // @ts-expect-error - extra properties should not be patchable
+            patchState(store, { nameStuff: 'b stuff' });
+          });
+          it('for named writables', async () => {
+            const Store = signalStore(
+              { providedIn: 'root', protectedState: false },
+              withResource(() => ({
+                name: restResourceWritable(() => 'a'),
+              })),
+            );
+
+            const store = TestBed.inject(Store);
+
+            await wait();
+
+            expect(store.nameStuff()).toBe('a stuff');
+
+            // @ts-expect-error - extra properties should not be patchable
+            patchState(store, { nameStuff: 'b stuff' });
+          });
+        });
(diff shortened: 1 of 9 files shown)


libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -19,20 +21,83 @@ import {
   withProps,
 } from '@ngrx/signals';
 
+// TODO - add other specifics of HttpResourceRef (headers, statusCode, progress)
```
michael-small: Should this be done in this PR? Or a separate PR? 

I will already have a separate PR for exposing `snapshot` as we discussed in a different PR

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
```diff
@@ -609,7 +612,40 @@ describe('withResource', () => {
...
+            }),
+          ),
+        })),
+        withResource(() =>
+          resource({
+            loader: () => Promise.resolve(-1),
+            defaultValue: 0,
+          }),
+        ),
+        withMethods((store) => ({
```
michael-small, line 898: @rainerhahnekamp I couldn't exactly figure out how to test this without a store, so I went with this approach for now.

michael-small:
Draft again: need to make the `Resource` not patchable

libs/ngrx-toolkit/src/lib/with-resource.ts
rainerhahnekamp, line 456: couldn't we reuse here `isResource` and additional check if value is writable and if we have reload?

libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -310,10 +434,44 @@ export function isResourceRef(value: unknown): value is ResourceRef<unknown> {
     'status' in value &&
     'error' in value &&
     'isLoading' in value &&
+    'snapshot' in value &&
     'hasValue' in value &&
     'reload' in value
   );
 }
+export function isHttpResourceRef(
```
rainerhahnekamp: reuse `isResourceRef`

libs/ngrx-toolkit/src/lib/with-resource.ts
rainerhahnekamp: I would really prefer is we could this an a separate, as I would also see this is a more generic feature. For example it should also allow to expose resource types which we aren't even aware of.

libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -19,46 +21,135 @@ import {
...
+    error: Signal<Error | undefined>;
+    isLoading: Signal<boolean>;
+    snapshot: Signal<ResourceSnapshot<InferResourceValue<T>>>;
+  };
+  methods: {
+    hasValue(): this is Resource<Exclude<InferResourceValue<T>, undefined>>;
+  } & ConditionalReloadMethod<T>;
+};
+
+type WidenedResource<T> = ResourceRef<T> | Resource<T> | HttpResourceRef<T>;
```
rainerhahnekamp: no httpresource. separate PR

libs/ngrx-toolkit/src/lib/with-resource.spec.ts
rainerhahnekamp: 👍  all good

rainerhahnekamp (changes requested):
I think we should the generic resource feature (see comments) in a separate issue

michael-small:
@rainerhahnekamp I have to make some changes so that `withEntityResource` can access the `value`. For now I am once again making this a draft so that it is reviewable as-is

michael-small:
@rainerhahnekamp 

Provided the following method in a new branch for `withEntityResource` is fine, then this PR for just `withResource` should be fine for a full review again: https://github.com/angular-architects/ngrx-toolkit/commit/12ab13f7fde79f2df4f662d4becf6f72bf386128#diff-77761f53f54f331f4b8c0fbd69af8150a1a1a790ac0ccc553522b6da66ffef4bR155

libs/ngrx-toolkit/src/lib/with-resource.ts
```diff
@@ -310,11 +461,69 @@ export function isResourceRef(value: unknown): value is ResourceRef<unknown> {
...
     'error' in value &&
     'isLoading' in value &&
-    'hasValue' in value &&
-    'reload' in value
+    'snapshot' in value &&
+    'hasValue' in value
+  );
+}
+
+export function isResourceRef(value: unknown): value is ResourceRef<unknown> {
```
msmallest, line 469: Refer to https://github.com/angular/angular/pull/70616/changes#diff-16f6b7cc4c4e69b511c8312d38c6cf71c412e6f1d6ac9092054c0f349215cdedR117 and the existing `isResource` for a better implementation, pending review

michael-small:
The state of the feature is ready for review. That said, I am adding tests for the extensions branch off of this to need to pass the type assertion, while still resolving to the right values which are applicable to this branch on its own.

PR #302, opened 2026-05-07
feat(devtools): expose tracked event payloads from withTrackedReducer

### LINKED ISSUE #294: RFC: Expose Event Payload in Redux DevTools via `withTrackedReducer`

sqrter:
`withTrackedReducer` now forwards full event objects to updateState so Redux DevTools can show payload data, not only action names.

Closes https://github.com/angular-architects/ngrx-toolkit/issues/294

### DIFF
--- a/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
@@ -4,7 +4,6 @@ import {
   WritableStateSource,
 } from '@ngrx/signals';
 import { currentActionNames } from './internal/current-action-names';
-import { Action } from './internal/models';
 
 type PatchFn = typeof originalPatchState extends (
   arg1: infer First,
@@ -29,7 +28,7 @@ export const patchState: PatchFn = (state, action, ...rest) => {
  */
 export function updateState<State extends object>(
   stateSource: WritableStateSource<State>,
-  action: string | Action,
+  action: string,
   ...updaters: Array<
     Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
   >
--- a/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
@@ -4,6 +4,7 @@ import {
   WritableStateSource,
 } from '@ngrx/signals';
 import { currentActionNames } from './internal/current-action-names';
+import { Action } from './internal/models';
 
 type PatchFn = typeof originalPatchState extends (
   arg1: infer First,
@@ -28,7 +29,7 @@ export const patchState: PatchFn = (state, action, ...rest) => {
  */
 export function updateState<State extends object>(
   stateSource: WritableStateSource<State>,
-  action: string,
+  action: string | Action,
   ...updaters: Array<
     Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
   >
--- a/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/update-state.ts
@@ -4,6 +4,7 @@ import {
   WritableStateSource,
 } from '@ngrx/signals';
 import { currentActionNames } from './internal/current-action-names';
+import { Action } from './internal/models';
 
 type PatchFn = typeof originalPatchState extends (
   arg1: infer First,
@@ -20,6 +21,14 @@ export const patchState: PatchFn = (state, action, ...rest) => {
 };
 
 /**
+ * Casts an object with a `type` property to an {@link Action}.
+ * Use this when you need to pass a structured action object to {@link updateState}.
+ */
+export function asAction<T extends { type: string }>(value: T): Action {
+  return value as unknown as Action;
+}
+
+/**
  * Wrapper of `patchState` for DevTools integration. Next to updating the state,
  * it also sends the action to the DevTools.
  * @param stateSource state of Signal Store
@@ -28,7 +37,7 @@ export const patchState: PatchFn = (state, action, ...rest) => {
  */
 export function updateState<State extends object>(
   stateSource: WritableStateSource<State>,
-  action: string,
+  action: string | Action,
   ...updaters: Array<
     Partial<NoInfer<State>> | PartialStateUpdater<NoInfer<State>>
   >
--- a/libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
+++ b/libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
@@ -3,7 +3,6 @@ import {
   EmptyFeatureResult,
   getState,
   PartialStateUpdater,
-  patchState,
   SignalStoreFeature,
   signalStoreFeature,
   type,
@@ -16,7 +15,7 @@ import {
 } from '@ngrx/signals/events';
 import { tap } from 'rxjs/operators';
 import { GLITCH_TRACKING_FEATURE } from './features/with-glitch-tracking';
-import { currentActionNames } from './internal/current-action-names';
+import { updateState, asAction } from './update-state';
 import { DEVTOOL_FEATURE_NAMES } from './with-devtools';
 
 export function withTrackedReducer<State extends object>(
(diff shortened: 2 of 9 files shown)


rainerhahnekamp:
Please let's not start multiple PRs for the same topic. I am closing that one. If there is no progress in #295 in the next 7 days, you can continue there.

sqrter:
Hi @rainerhahnekamp, thanks for your response. I created it separately simply because I didn’t want to interfere with the changes that @OleksandrOleniuk made. Besides, I wasn’t sure if I understood your comments correctly. From my perspective, it would be better to continue the discussion on the newer pull request, since the old one hasn’t had any activity for a while + changing it means more work to do than just moving on with the new one.

rainerhahnekamp:
@sqrter I've pinged @OleksandrOleniuk. Let's give him the 7 days

libs/ngrx-toolkit/src/lib/devtools/update-state.ts
```diff
@@ -28,7 +29,7 @@ export const patchState: PatchFn = (state, action, ...rest) => {
  */
 export function updateState<State extends object>(
   stateSource: WritableStateSource<State>,
-  action: string,
+  action: string | Action,
```
rainerhahnekamp, line 40: careful, that could be a breaking change. because if someone has a state of type `{type:  string}`, TypeScript could have an issue in detecting if this is now the action name or update.
so better to pass on `type` in `with-tracked-reducer` and keep `updateState` untouched.

rainerhahnekamp (changes requested):
Thanks, great work, just one comment.

libs/ngrx-toolkit/src/lib/devtools/update-state.ts
sqrter: @rainerhahnekamp, thanks for the review — you're absolutely right about the breaking change risk.
I reverted updateState to action: string only, keeping the public API untouched. To still forward event payloads from withTrackedReducer, I replaced the updateState call inside the reducer with direct currentActionNames.add(event) + patchState(store, ...updaters). This way withTrackedReducer keeps showing full payloads in DevTools, but updateState stays backward compatible.

libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
```diff
@@ -39,7 +40,8 @@ export function withTrackedReducer<State extends object>(
             const result = caseReducer.reducer(event, state);
             const updaters = Array.isArray(result) ? result : [result];
 
-            updateState(store, event.type, ...updaters);
+            currentActionNames.add(event);
```
rainerhahnekamp: i would really like to avoid accessing that one outside the core. are there other options?

I was thinking that we could maybe try out a branded type for action. then we could keep the action as a whole.

what do you think?

rainerhahnekamp (changes requested):

libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
sqrter: Ok, I see, but I'm not sure whether a branded type wouldn't be an overkill, simply because the users who want to manually pass Action objects would need an explicit as Action cast or similar... I think we could just completely revert the changes in `with-tracked-reducer.ts`, and in `updateState`, the action type would again become `action: string | Action,`. If this tiny type change in the `updateState` is acceptable, I can do this.

sqrter: @rainerhahnekamp Looking back at the conversation, I realized that it's the point about the breaking change you mentioned earlier. As far as I undertant, the edge case is when `State = { type: string }` and the user would call `updateState(store, {type:'someType', payload: 'somePayload'}, { type: 'bar' })`, right? 

So the question is, would it be fine if we:

1. create a branded type for action, something like:

```typescript
declare const __actionBrand: unique symbol;
export type Action = { type: string; [key: string]: unknown } & {
  readonly [__actionBrand]: true;
};
```
2. export it from public API

The usage would be like `updateState(store, { type: 'Load', id: '1' } as Action, { loading: true });`

Or we could also have some helper function `updateState(store, createAction('[Books] Load', { id: '1' }), { loading: true });`

What do you think?

sqrter: UPD: I updated the pull request to roughly see what it might look like. Still, I added a function `asAction` so we don't have to cast like `as unknown as Action` everywhere. I haven't added anything to the docs yet.

sqrter: Hi @rainerhahnekamp, have you had a chance to review the version with "branded type" implementation? Just let me know if there's anything more I need to adjust. Thanks

sqrter: Hey, @rainerhahnekamp, apologies for the ping — just checking if you're still able to review this? Would love to get it merged. Thanks!

rainerhahnekamp (approved):

libs/ngrx-toolkit/src/lib/devtools/with-tracked-reducer.ts
rainerhahnekamp: @sqrter Thanks for the reminder. Has been merged and will be available in the next version, which is when NgRx 22 is released.

PR #303, opened 2026-05-13
refactor: add types for mutation errors

### LINKED ISSUE #285: mutations: specifying the type for errors
Currently, the type for mutation errors is just `unknown`. So whether you are awaiting a mutation or just accessing the error in `onError`, you must cast or narrow your error if its type is knowable.

This issue is to track adding an error typing functionality. I have picked this up formally, but feel free to discuss.

To avoid breaking changes, we are thinking of adding an `inferError: boolean` which would be false for all versions `<=21`. And then in 22, it would default to `true`.

michael-small:
Closes #285

### DIFF
--- a/apps/demo/src/app/counter-rx-mutation/counter-rx-mutation.ts
+++ b/apps/demo/src/app/counter-rx-mutation/counter-rx-mutation.ts
@@ -5,7 +5,7 @@ import {
 } from '@angular-architects/ngrx-toolkit';
 import { CommonModule } from '@angular/common';
 import { Component, computed, signal } from '@angular/core';
-import { delay, Observable, of, throwError } from 'rxjs';
+import { catchError, delay, Observable, of, throwError } from 'rxjs';
 
 export type Params = {
   value: number;
@@ -51,6 +51,12 @@ export class CounterRxMutation {
     onSuccess: (response) => {
       console.log('Counter sent to server:', response);
     },
+    // TODO - In the current state, parsing for the error does not narrow unless either
+    // - 3rd error type generic specified
+    // - No generic types specified
+    // This is inline with the existing configuration of this `httpMutation` example before this proposed change
+    // If you uncomment the following, you must still specify the error type or drop the explicit types
+    // parseError: (error) => error as string,
     onError: (error) => {
       console.error('Failed to send counter:', error);
     },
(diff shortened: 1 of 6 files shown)


apps/demo/src/app/counter-rx-mutation/counter-rx-mutation.ts
```diff
@@ -51,6 +51,12 @@ export class CounterRxMutation {
     onSuccess: (response) => {
       console.log('Counter sent to server:', response);
     },
+    // TODO - In the current state, parsing for the error does not narrow unless either
```
michael-small, line 54: Provided this is the behavior we think makes sense, then I just need to add a test at this point. Thoughts?

michael-small: @rainerhahnekamp we spoke about this privately a few months back, and you proposed a flag, but I think this solution is sufficient and IMO better. Wondering before I just write up some tests/docs on this.

Sound good?

rainerhahnekamp: Yes that definitely makes sense

libs/ngrx-toolkit/src/lib/mutation/mutation.ts
```diff
@@ -1,26 +1,26 @@
 import { Signal } from '@angular/core';
 
-export type MutationResult<Result> =
+export type MutationResult<Result, Err> =
```
rainerhahnekamp: if that type is public api we need a default value for `Err` as well. Otherwise it would be breaking

libs/ngrx-toolkit/src/lib/mutation/mutation.ts
```diff
@@ -1,26 +1,26 @@
...
     }
   | {
       status: 'aborted';
     };
 
 export type MutationStatus = 'idle' | 'pending' | 'error' | 'success';
 
-export type Mutation<Parameter, Result> = {
-  (params: Parameter): Promise<MutationResult<Result>>;
+export type Mutation<Parameter, Result, Err> = {
```
rainerhahnekamp: also here default type if part of the public API

rainerhahnekamp (changes requested):

PR #304, opened 2026-05-23
feat(devtools): create tracked event and tracked event group using proxy

ahmedwelhakim:
# Proposal: Add `trackedEvent` and `trackedEventGroup`

## Motivation

While working with event-driven patterns in NgRx Signal Store, debugging and tracing dispatched events can become difficult as the application grows.
As withTrackedReducer doesn't detect all dispatched events, only used ones in reducer.

So this proposal introduces:

* `trackedEvent` to be used instead of `event`
* `trackedEventGroup` to be used instead of `eventGroup`

as lightweight wrappers around the existing event creators to track dispatched events using proxy without needing the withTrackedReducer.

---

## Goals

* Preserve the exact same API as the original event creators
* Add transparent event dispatch tracking
* Keep typings fully compatible with existing NgRx Signal Store patterns
* Avoid runtime overhead unless events are actually dispatched

---

### `trackedEvent`

```ts
export const loadBooks = trackedEvent('[Books] Load Books');
```

```ts
loadBooks();
```

Devtool action:

```txt
[Books] Load Books
```

---

### `trackedEventGroup`

```ts
export const bookEvents = trackedEventGroup({
  source: 'Book Store',
  events: {
    loadBooks: type<void>(),
    bookSelected: type<{ bookId: string }>(),
  },
});
```

```ts
bookEvents.bookSelected({ bookId: '1' });
```

Devtool:

```txt
[Book Store] bookSelected
```

### DIFF
--- a/libs/ngrx-toolkit/src/index.ts
+++ b/libs/ngrx-toolkit/src/index.ts
@@ -6,11 +6,12 @@ export {
   provideDevtoolsConfig,
 } from './lib/devtools/provide-devtools-config';
 export { renameDevtoolsName } from './lib/devtools/rename-devtools-name';
+export { trackedEvent } from './lib/devtools/tracked-event';
+export { trackedEventGroup } from './lib/devtools/tracked-event-group';
 export { patchState, updateState } from './lib/devtools/update-state';
 export { withDevToolsStub } from './lib/devtools/with-dev-tools-stub';
 export { DevtoolsFeature, withDevtools } from './lib/devtools/with-devtools';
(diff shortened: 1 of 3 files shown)


rainerhahnekamp:
The issue here is that devtools focuses on the SignalStore instances, because that's where the state lives. If devtools listened for events and not `watchState`, we would have to enable them globally for all stores. At the moment, I don't know how to do that. It would require changes to `signalStore` itself.

I do get your point that unhandled events might not show up, though. We are currently working on a redesign of the devtools which would land in `@ngrx/signals`. Can we revisit your issue once that design is done?

PR #305, opened 2026-05-25
fix: narrow `NamedMutationMethods` to just fn signature

### LINKED ISSUE #286: withMutation Typing
```ts
// store
withMutations((store) => ({
  myMutation: httpMutation((myNum: number) => ({
    url: `/api/register/myMutation/${myNum}`,
    method: 'GET',
    headers: apiHeaders(),
  })),
})),
```

wrong usage (example 1)
```ts
// component
readonly myMutation = this.store.myMutation; // Mutation<number, any>

// html
{{ myMutation.isPending() }}

// runtime error:
ERROR TypeError: ctx.myMutation.isPending is not a function
```

correct usage (example 2)
```ts
// component
readonly myMutationIsPending = this.store.myMutationIsPending;

// html
{{ myMutationIsPending() }}
```
- I would love that `example 1` would work. 
- The Typings give you the feeling that `example 1` should work, but then you end in a runtime error.

michael-small:
Closes #286

### DIFF
--- a/libs/ngrx-toolkit/src/lib/with-mutations.ts
+++ b/libs/ngrx-toolkit/src/lib/with-mutations.ts
@@ -10,7 +10,7 @@ import {
   withMethods,
   WritableStateSource,
 } from '@ngrx/signals';
-import { Mutation, MutationStatus } from './mutation/mutation';
+import { Mutation, MutationResult, MutationStatus } from './mutation/mutation';
 
 // NamedMutationMethods below will infer the actual parameter and return types
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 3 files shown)


libs/ngrx-toolkit/src/lib/with-mutations.ts
```diff
@@ -33,7 +33,7 @@ type NamedMutationMethods<T extends MutationsDictionary> = {
     infer P,
     infer R
   >
-    ? Mutation<P, R>
+    ? (params: P) => Promise<MutationResult<R>>
```
rainerhahnekamp, line 36: do we have a test for that?

rainerhahnekamp (changes requested):

libs/ngrx-toolkit/src/lib/with-mutations.ts
michael-small: I made `does not expose mutation properties directly, only the method` for this, should it be more specific?

michael-small: I added a more explicit test

rainerhahnekamp (approved):
👍

PR #315, opened 2026-08-27
fix(with-resource): preserve hasValue narrowing

rainerhahnekamp:
## Summary
- model `hasValue` as a function property with an explicit generic receiver
- preserve the concrete SignalStore type while narrowing `value()` to exclude `undefined`
- turn the existing inferred-consumer type tests into positive narrowing assertions

## Verification
- full pre-commit `verify:all` (lint, test, and build for the library and demo)
- focused TypeScript compilation confirms the three `hasValue` narrowing assertions; one unrelated existing `AddressResolver.address` fixture error remains

### DIFF
--- a/apps/demo/src/app/with-resource/with-resource.component.html
+++ b/apps/demo/src/app/with-resource/with-resource.component.html
@@ -1,0 +1,19 @@
+,
+<h1>withResource</h1>
+<a
+  href="https://ngrx-toolkit.angulararchitects.io/docs/with-resource"
+  target="_blank"
+  >withResource doc page</a
+>
+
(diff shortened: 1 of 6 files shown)


michael-small (changes requested):
This change seems great for unnamed resources, but it does not fix named resources.

Repro branch commit, with examples you can intelisense in a component HTML/TS as well as a purposely broken test: https://github.com/ngrx-toolkit/ngrx-toolkit/commit/5b8cb3486fb4804aa1122b15d053fb7372f00ed1

michael-small (approved):
LGTM

